PR #18, opened 2015-03-30
fixes #5098, properly detect has one relations

nineinchnick:
Fixes yiisoft/yii2#5098. Properly compare primary and all unique keys to foreign key columns.
Added unit tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,18 @@
+# phpstorm project files
+.idea
+
+# netbeans project files
+nbproject
+
+# zend studio for eclipse project files
+.buildpath
+.project
+.settings
+
+# windows thumbnail cache
+Thumbs.db
+
+# composer vendor dir
 /vendor
 /composer.lock
 /tests/runtime
--- a/.gitignore
+++ b/.gitignore
@@ -1,18 +1,3 @@
-# phpstorm project files
-.idea
-
-# netbeans project files
-nbproject
-
-# zend studio for eclipse project files
-.buildpath
-.project
-.settings
-
-# windows thumbnail cache
-Thumbs.db
-
-# composer vendor dir
 /vendor
 /composer.lock
 /tests/runtime
--- a/generators/model/Generator.php
+++ b/generators/model/Generator.php
@@ -323,15 +323,17 @@ class Generator extends \yii\gii\Generator
                 ];
 
                 // Add relation for the referenced table
-                $hasMany = false;
-                if (count($table->primaryKey) > count($fks)) {
-                    $hasMany = true;
-                } else {
-                    foreach ($fks as $key) {
-                        if (!in_array($key, $table->primaryKey, true)) {
-                            $hasMany = true;
-                            break;
-                        }
+                $uniqueKeys = [$table->primaryKey];
+                try {
+                    $uniqueKeys = array_merge($uniqueKeys, $db->getSchema()->findUniqueIndexes($table));
+                } catch (NotSupportedException $e) {
+                    // pass
+                }
+                $hasMany = true;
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -1,3 +1,18 @@
...
+nbproject
+
+# zend studio for eclipse project files
+.buildpath
+.project
+.settings
+
+# windows thumbnail cache
+Thumbs.db
+
```
cebe: this is not needed, you should have a gitignore global file for such things.
see https://gist.github.com/cebe/1166792#file-gitignore_global for an example.

generators/model/Generator.php
```diff
@@ -323,15 +323,17 @@ protected function generateRelations()
...
-                    foreach ($fks as $key) {
-                        if (!in_array($key, $table->primaryKey, true)) {
-                            $hasMany = true;
-                            break;
-                        }
+                $uniqueKeys = [$table->primaryKey];
+                try {
+                    $uniqueKeys = array_merge($uniqueKeys, $db->getSchema()->findUniqueIndexes($table));
+                } catch (NotSupportedException $e) {
+                    // pass
```
samdark: You mean ignore?

samdark:
Looks good to me. Need changelog.

nineinchnick:
@samdark this one's ready for merging.

samdark:
Merged https://github.com/yiisoft/yii2-gii/commit/00719bfd11760c8f442b87ff24b858339e753bfe. Thanks!

WadeShuler:
I think Gii is still generating the wrong relation.. Wanted to mention it here before opening a new issue, just to be sure it isn't something I should have done differently.

I used the include init migration to create the user table. I then created a new migration to create a user_profile table.

    $this->createTable('{{%user_profile}}', [
            'id' => $this->primaryKey(),
            'user_id' => $this->integer()->notNull(),
            'nicename' => $this->string(30)->notNull(),
            'use_gravatar' => $this->boolean()->defaultValue(1),
            'receive_emails' => $this->boolean()->defaultValue(1),
            'share_details' => $this->boolean()->defaultValue(1),

            'phone' => $this->string(50)->notNull(),
            'address1' => $this->string(100)->notNull(),
            'address2' => $this->string(100)->notNull(),
            'city' => $this->string(100)->notNull(),
            'state' => $this->string(100)->notNull(),
            'zip' => $this->string(20)->notNull(),
            'country' => $this->string(100)->notNull(),

            'created_at' => $this->integer()->notNull(),
            'updated_at' => $this->integer()->notNull(),
        ], $tableOptions);
        $this->addForeignKey('fk_user_profile_user_id', '{{%user_profile}}', 'user_id', '{{%user}}', 'id', 'CASCADE', 'CASCADE');

As you can see, it creates a foreign key mapping `user_profiles.user_id` to `user.id`. Nothing fancy there :)

However, Gii generated a `hasMany()` relation for it...

Since `user.id` is a primaryKey, Gii should recognize this and generate a `hasOne` relation, shouldn't it?

samdark:
The structure reads as User hasMany profiles. Profile hasOne User. I think there's no way to detect that in your case there's only one profile.

nineinchnick:
You have to have an unique key on the _foreign key columns_ for the relation to be hasOne. A primary key is also unique. And you can't create a foreign key pointing to a non-unique column at all.

WadeShuler:
What if I added the `unique()` function to the integer columns that have relations?

ie:

    'user_id' => $this->integer()->notNull()->unique(),

Should it create a hasOne() relation then?

nineinchnick:
@WadeShuler yes, that's what I've said.

PR #22, opened 2015-04-06
Update model.php to generate const properties for Model's attributes

mnglkhn:
Generate const values for model's property names. 
Prefixed by 'COL_', can then be accessed like Model::COL_PROP_NAME.

### DIFF
--- a/generators/model/default/model.php
+++ b/generators/model/default/model.php
@@ -34,6 +34,11 @@ use Yii;
  */
 class <?= $className ?> extends <?= '\\' . ltrim($generator->baseClass, '\\') . "\n" ?>
 {
+
+<?php foreach ($tableSchema->columns as $column): ?>
+    const COL_<?= strtoupper($column->name) ?> = '<?= $column->name ?>';
+<?php endforeach; ?>
+
     /**
      * @inheritdoc
      */


samdark:
Why do we need it?

mnglkhn:
When working with an IDE, such as PhpStorm, allows for quick model introspection, without the need to guess what the actual field names are. Especially when you fetch a model as an array. At that point you need to know the array key names, in order to get the values. Cuts down on errors.
I just type Model::COL_ and then CTRL+SPACE and get all the Model columns.

samdark:
If you're working with objects, modern IDEs are already using `@property` declarations and autocompleting properties.

mnglkhn:
hmm, I might be missing something and I am curious how would that work in the following case:

$category = Category->find()->where([...])->toArray();
echo $category['label'];

which would become:
echo $category[Category::COL_LABEL];

For some reason, I find the current way prone to errors. Plus you would need to jump to the model definition to see the actual column name and then copy it in order to correctly paste it back where you need it in your workflow.

samdark:
It won't work with arrays, of course.

samdark:
```php
$category = Category->find()->where([...])->one();
echo $category->label;
```

mnglkhn:
Of course, if you keep the model object, then you have access to its properties, no argument there.
My thinking was that somebody might want to move arrays around for increased speed and lower memory size usage.

samdark:
Well, that's not reall common case so I think it's better for you to keep it as custom template for now. Will reconsider if more people will ask for it.

PR #23, opened 2015-04-07
7830 multi schema support

nineinchnick:
Resolves yiisoft/yii2#7830. Adds support for relations between tables in different schemas.


generators/model/Generator.php
```diff
@@ -287,6 +290,45 @@ public function generateRules($table)
     }
 
     /**
+     * @param \yii\db\TableSchema the table being checked
```
samdark: Description of what method does is required,

generators/model/Generator.php
```diff
@@ -287,6 +290,45 @@ public function generateRules($table)
...
+     * @param array $fks
+     * @param array $relations
+     * @return array
+     */
+    private function generateManyManyRelations($table, $fks, $relations)
+    {
+        $db         = $this->getDbConnection();
```
samdark: Generally we aren't aligning assignments like it's done here.

generators/model/Generator.php
```diff
@@ -287,6 +290,45 @@ public function generateRules($table)
...
+        $table0     = $fks[$table->primaryKey[0]][0];
```
samdark: Could it be named better than `$table0`? Perharps `$pkPrimaryTable`?

generators/model/Generator.php
```diff
@@ -297,79 +339,65 @@ protected function generateRelations()
...
-                        $hasMany = false;
-                        break;
+        foreach ($schemaNames as $schemaName) {
+            foreach ($db->getSchema()->getTableSchemas($schemaName) as $table) {
+                $className = $this->generateClassName($table->fullName);
+                foreach ($table->foreignKeys as $refs) {
+                    $refTable = $refs[0];
+                    $refTableSchema = $db->getTableSchema($refTable);
+                    unset($refs[0]);
+                    $fks          = array_keys($refs);
```
samdark: Better not to align assignments. It's no consistent with overall code style.

generators/model/Generator.php
```diff
@@ -297,79 +339,65 @@ protected function generateRelations()
...
-
-            $link = $this->generateRelationLink([$fks[$table->primaryKey[1]][1] => $table->primaryKey[1]]);
-            $viaLink = $this->generateRelationLink([$table->primaryKey[0] => $fks[$table->primaryKey[0]][1]]);
-            $relationName = $this->generateRelationName($relations, $className0, $db->getTableSchema($table0), $table->primaryKey[1], true);
-            $relations[$className0][$relationName] = [
-                "return \$this->hasMany($className1::className(), $link)->viaTable('" . $this->generateTableName($table->name) . "', $viaLink);",
-                $className1,
-                true,
-            ];
+                if (($fks = $this->checkPivotTable($table)) === false) {
```
samdark: As I remember, correct name for the "middle" table is "junction table", not "pivot table". The latter is something else.

generators/model/Generator.php
nineinchnick: This is the old code, I've just moved it from `generateRelations()` to a separate method. I've got another issue created about this so I'm going to refactor it later.

generators/model/Generator.php
nineinchnick: Again, I just changed the indentation here, the naming remains. I also like "junction table" better but I don't want to introduce unecessary changes.

generators/model/Generator.php
samdark: OK.

generators/model/Generator.php
samdark: OK.

nineinchnick:
Oh, forgot to mention that this PR requires yiisoft/yii#8012 to be accepted first, that's why builds fail now.

samdark:
8012 merged.

samdark:
I think it's not the reason for build failure: https://travis-ci.org/yiisoft/yii2-gii/jobs/57717304#L205. See the message that SQLite doesn't have anything about schema...

nineinchnick:
Tests run against yii that's in composer.json and it's 2.0.3. Also yiisoft/yii2#7973 from upcoming 2.0.4 is required here.

samdark:
Right.

samdark:
Yep, verified locally that tests are OK.

samdark:
Thank you!

PR #41, opened 2015-05-20
Sort the labels alphabetically

thiagotalma:
In order to facilitate the location of the labels when we have models with a large number of attributes.

### DIFF
--- a/generators/model/Generator.php
+++ b/generators/model/Generator.php
@@ -237,6 +237,7 @@ class Generator extends \yii\gii\Generator
                 $labels[$column->name] = $label;
             }
         }
+        ksort($labels);
 
         return $labels;
     }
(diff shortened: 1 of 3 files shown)


klimov-paul:
I don't think such change should be made. Curretly labels are generated in the same order attributes are fetched from database. Usually people prefer them in this way.

For example: assume I have a table 'Item', which has columns: 'id', 'active', 'name'. I prefer to have the attributes listed in the order I gave them, while creating table: 'id', 'active', 'name'. But with this change the order will be following: 'active', 'id', 'name'.

Although, if such change applied, same sorting should be used, while composing `@property` PHP docs.

thiagotalma:
Hello @klimov-paul ,

It really is not a good option to force the sort.
Nothing is better to give the option to the developer.

ruffiem:
It depends on the developer. For me, having the cloning the table structure makes more sense. Tho, I understand your point of view @thiagotalma

samdark:
Being an option it complicates the form while rare developer will prefer ID as non-first field. Thanks for proposal but it won't be merged.

PR #42, opened 2015-05-21
Facilitate copy of the code

thiagotalma:
It is much more convenient than having to select with the mouse throughout.

### DIFF
--- a/assets/main.css
+++ b/assets/main.css
@@ -250,3 +250,18 @@ body {
     margin: 0;
 }
 
+#clipboard-container {
+    position: fixed;
+    left: 0;
+    top: 0;
+    width: 0;
+    height: 0;
+    z-index: 100;
+    /*display: none;*/
+    opacity: 0;
+}
+#clipboard {
+    width: 1px;
+    height: 1px;
+    padding: 0;
+}
--- a/assets/gii.js
+++ b/assets/gii.js
@@ -78,7 +78,11 @@ yii.gii = (function ($) {
             $modal.find('.modal-next').attr('href', $next.attr('href')).data('title', $next.data('title'));
         }
         $modalBody.html(data);
-        valueToCopy = $modalBody.get(0).innerText;
+
+        var tmp = document.createElement("div");
+        tmp.innerHTML = data;
+        valueToCopy = tmp.innerText || tmp.textContent || "";
+
         $modal.find('.content').css('max-height', ($(window).height() - 200) + 'px');
     };
 
--- a/assets/gii.js
+++ b/assets/gii.js
@@ -1,5 +1,34 @@
 yii.gii = (function ($) {
-    var isActive = $('.default-view').length > 0;
+
+    var $clipboardContainer = $("#clipboard-container"),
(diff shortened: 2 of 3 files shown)


samdark:
Not sure it's the best way to do it.

qip shot - screen 246

qip shot - screen 247

How about selecting the text itself instead of providing it in an input box?

thiagotalma:
The input is foolproof, works on any browser.

If you have any other suggestions just tell me that I change the PR.

samdark:
Here's very smart way Trello handles it: https://stackoverflow.com/questions/17527870/how-does-trello-access-the-users-clipboard. So we can probably do it the same way removing any buttons and input dialogs and showing "Press CTRL+C to copy" instead.

thiagotalma:
@samdark, very good!

assets/main.css
```diff
@@ -250,3 +250,18 @@ body {
...
 }
 
+#clipboard-container {
+    position: fixed;
+    left: 0;
+    top: 0;
+    width: 0;
+    height: 0;
+    z-index: 100;
+    /*display: none;*/
```
samdark: Remove?

samdark:
Overall works but it adds unnecessary newlines:

```php

<?php


namespace app\models;



use Yii;



/**

 * This is the model class for table "user".

 *

 * @property integer $id
 * @property string $username
 * @property string $auth_key
```

assets/main.css
thiagotalma: Nop! Its required.

thiagotalma:
Strange .... here works very well without extra spaces. (remember my first answer I gave you this PR?)

thiagotalma:
```PHP
<?php

namespace common\models;

use Yii;

/**
 * This is the model class for table "tbl_user".
 *
 * @property integer $id
 * @property string $username
 * @property string $email
 * @property string $password_hash
 * @property string $auth_key
 * @property integer $confirmed_at
 */
class User extends \yii\db\ActiveRecord
{
    /**
     * @inheritdoc
     */
    public static function tableName()
    {
        return 'tbl_user';
    }

    /**
     * @inheritdoc
     */
    public function rules()
    {
    }
```

samdark:
That's Chrome on windows 8. Probably browser differences...

thiagotalma:
@samdark, I can not simulate here. You can help me test, please?

See now worked.

samdark:
Merged with some improvements.

qip shot - screen 249

prabowomurti:
I am using Cmd + C and Chrome. But the copied character of space become `&nbsp;` character. So I always get the syntax error message. Do you have any idea why this happens?

cebe:
@prabowomurti could you please open a new issue for this?

thiagotalma:
In fact, it never worked.
I ended up giving a specific solution for my needs, creating a macro using Autohotkey.

cebe:
@thiagotalma I've been using it a few times with firefox and had no issue with it.

thiagotalma:
@cebe So it's just a problem with Chrome...

PR #51, opened 2015-05-30
ActiveRecord class moved into the header

LAV45:
This will generate
```php
use Yii;
use yii\db\ActiveRecord;

class Post extends ActiveRecord
{
...
}

### DIFF
--- a/generators/model/default/model.php
+++ b/generators/model/default/model.php
@@ -19,6 +19,7 @@ echo "<?php\n";
 namespace <?= $generator->ns ?>;
 
 use Yii;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by cebe** [comment]
This is intentional to not conflict with probably existing `ActiveRecord` class in the same namespace.

**Comment by LAV45** [comment]
What if to add validation?
```php
$compactBaseClass = substr(strrchr(ltrim($generator->baseClass, '\\'), "\\"), 1);
if ($className == $compactBaseClass) {
    $compactBaseClass = 'Base' . $compactBaseClass;
}
```

Issue #56, opened 2015-06-13
generate 'exist' rules based on the foreign keys on the table.


Faryshta:
i think the title is self explanatory. It can be very useful specially to avoid exceptions being thrown to the end user when he tries to add a value from an unexistant relation.

samdark:
Yes, that would be great.

PR #58, opened 2015-06-14
Generate 'exist' rules based on foreign keys

Faryshta:


Faryshta:
@samdark #56

generators/model/Generator.php
```diff
@@ -317,6 +317,20 @@ public function generateRules($table)
             // doesn't support unique indexes information...do nothing
         }
 
+        // Exist rules for foreign keys.
+        foreach ($table->foreignKeys as $refs) {
+            $refClassName = $this->generateClassName($refs[0]);
+            unset($refs[0]);
+            $rules[] = "[['" . implode("', '", array_keys($refs)) . "'], "
+                . "'exist', "
+                . "'skypOnError' => true, "
```
samdark: skyp → skip

Faryshta:
what do I need to do to have this shipped with 2.0.5? @samdark ?

cebe:
looks good to me.

samdark:
Adjusted code https://github.com/yiisoft/yii2-gii/commit/4487b4600525f04b5e2e8a46593f04f62a9c3446 and merged. @Faryshta thanks!

thiagotalma:
@samdark @cebe @Faryshta 
Guys, you need to check the need to use the brackets?

Here there was an error and found that the reason was the use of brackets when there is only one attribute.

The attribute is compared to NULL.

samdark:
@thiagotalma can you elaborate?

thiagotalma:
Generated rule:

```php
public function rules()
{
    return [
        [['client_id'], 'exist', 'skipOnError' => true, 'targetClass' => Client::className(), 'targetAttribute' => ['id']],
    ];
}
```

Query:
```sql
SELECT 1 FROM `pvw_Client` WHERE `id` IS NULL
```

Fixed rule:


```php
public function rules()
{
    return [
        ['client_id', 'exist', 'skipOnError' => true, 'targetClass' => Client::className(), 'targetAttribute' => 'id'],
    ];
}
```

Query:
```sql
SELECT 1 FROM `pvw_Client` WHERE `id`=135
```

samdark:
https://github.com/yiisoft/yii2-gii/issues/64 ?

Faryshta:
would

```php
public function rules()
{
    return [
        ['client_id', 'exist', 'skipOnError' => true, 'targetClass' => Client::className(), 'targetAttribute' => ['client_id' => 'id']],
    ];
}
```

work?

thiagotalma:
@Faryshta same result (correct)

```sql
SELECT 1 FROM `pvw_Client` WHERE `id`=135
```

@samdark, not the same issue.

samdark:
@thiagotalma, @Faryshta would you like to fix it?

thiagotalma:
@Faryshta its your baby 😉

cebe:
@thiagotalma can you open a new issue for that so we do not forget about it?

Faryshta:
@thiagotalma @samdark @cebe i will fix it

Faryshta:
#68 fix it and it passes the tests now.

PR #60, opened 2015-06-15
Doc snippet for the module generator

Faryshta:


generators/module/default/module.php
```diff
@@ -16,10 +16,19 @@
 
 namespace <?= $ns ?>;
 
+/**
+ * Class to initialize the module `<?= $generator->moduleID ?>`
```
klimov-paul: 'Class to initialize the module' does not sounds appropriate.
Docs for the module class usually contains description of its purpose, like 'module for user private messages management' or something.

Faryshta: what should i put there then?

Faryshta:
any update on this? @klimov-paul @SilverFire ?

generators/module/default/controller.php
```diff
@@ -13,8 +13,14 @@
 
 use yii\web\Controller;
 
+/**
+ * Default controller for the `<?= $generator->moduleID ?>` module
+ */
 class DefaultController extends Controller
 {
+    /**
+     * render the default view for the module
```
SilverFire: `Render` -> `Renders`

SilverFire: `default` -> `index`

generators/module/default/module.php
SilverFire: What about

`<?= $generator->moduleID ?> module definition class`

Faryshta:
proposed changes applied

SilverFire:
Merged, thank you!

PR #62, opened 2015-07-01
Option to define path to save model files

thiagotalma:
The folder where the files are saved are hardcoded.

This PR adds a field to let the developer choose the folder where to save the files.

By default, the folder is defined as before, using the namespace alias.

A use case: create snapshots of the unchanged original model files.


klimov-paul:
Folder is not harcoded is is composed from model class namespace.
Such addition does not make much sense, while it overcomplecate things.

PR #63, opened 2015-07-09
add comma after last line in rules

nkovacs:
This is more convenient.
attributeLabels and the rules in the crud generator's search model (generators/crud/default/search.php) already has this.


samdark:
Thanks!

PR #69, opened 2015-07-25
Gii overwrites your file even when you tell it not to

chernyshev:
Related issue: https://github.com/yiisoft/yii2/issues/9163

### DIFF
--- a/Generator.php
+++ b/Generator.php
@@ -270,7 +270,7 @@ abstract class Generator extends Model
         $hasError = false;
         foreach ($files as $file) {
             $relativePath = $file->getRelativePath();
-            if (isset($answers[$file->id]) && $file->operation !== CodeFile::OP_SKIP) {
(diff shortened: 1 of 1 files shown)


Generator.php
```diff
@@ -270,7 +270,7 @@ public function save($files, $answers, &$results)
         $hasError = false;
         foreach ($files as $file) {
             $relativePath = $file->getRelativePath();
-            if (isset($answers[$file->id]) && $file->operation !== CodeFile::OP_SKIP) {
+            if (isset($answers[$file->id]) && $answers[$file->id] === true && $file->operation !== CodeFile::OP_SKIP) {
```
jeicd: The ```$answers``` array in web controller is filling with string type values. String comparison brakes the check.

samdark: What if it's changed to `!empty($answers[$file->id])`?

jeicd: Yes, that will be fine

PR #70, opened 2015-07-26
[incomplete] generate the form attributes based on validations

Faryshta:
i haven't find the best strategy to generate the activ fields yet. @samdark any advice there?

### DIFF
--- a/generators/crud/default/views/_form.php
+++ b/generators/crud/default/views/_form.php
@@ -28,15 +28,48 @@ use yii\widgets\ActiveForm;
 
     <?= "<?php " ?>$form = ActiveForm::begin(); ?>
 
-<?php foreach ($generator->getColumnNames() as $attribute) {
-    if (in_array($attribute, $safeAttributes)) {
-        echo "    <?= " . $generator->generateActiveField($attribute) . " ?>\n\n";
+<?php 
+$weights = [
+    'yii\validators\ExistValidator' => 2,
+    'yii\validators\FileValidator' => 2,
+    'yii\validators\ImageValidator' => 2,
+    'yii\validators\CapchaValidator' => 2,
+    'yii\validators\DateValidator' => 1,
+    'yii\validators\RangeValidator' => 1,
+    'yii\validators\NumberValidator' => 1,
+    'yii\validators\BooleanValidator' => 0,
+    'yii\validators\EmailValidator' => 0,
+    'yii\validators\NumberValidator' => 0,
+];
+
+$model = new $generator->modelClass();
+$attributes = array_flip($model->safeAttributes());
+
+foreach ($model->validators as $validator) {
+    if (!isset($weights[get_class($validator)])) {
+        continue;
     }
-} ?>
+
+    $weight = $weights[get_class($validator)];
+    foreach ($validator->attributes as $attr) {
+        if (!is_array($attributes[$attr])
+            or $weight <= $attributes[$attr]['weight']
+        ) {
+            continue;
+        }
+
+        $attributes[$attr] => ['weight' => $weight, 'validator' => $validator];
+    }
+}
+
+foreach ($attributes as $attr => $value) {
+    // code to generate each attribute based on the validators saved sofar.
+}
+    ?>
     <div class="form-group">
         <?= "<?= " ?>Html::submitButton($model->isNewRecord ? <?= $generator->generateString('Create') ?> : <?= $generator->generateString('Update') ?>, ['class' => $model->isNewRecord ? 'btn btn-success' : 'btn btn-primary']) ?>
     </div>
-
+    
     <?= "<?php " ?>ActiveForm::end(); ?>
 
 </div>


Faryshta:
@samdark can we have this for 2.0.7?

cebe:
we may even have this for 2.0.5 ;) extensions are now released separately from frameworks.

Faryshta:
@cebe what do you think on https://github.com/yiisoft/yii2-gii/pull/70/files#diff-4704dd066f5268da66c24cb6ef42b0e3R66 ?

Should we make a huge `switch` with a case for every possible type of validator?

generators/crud/default/views/_form.php
```diff
@@ -28,15 +28,48 @@
 
     <?= "<?php " ?>$form = ActiveForm::begin(); ?>
 
-<?php foreach ($generator->getColumnNames() as $attribute) {
-    if (in_array($attribute, $safeAttributes)) {
-        echo "    <?= " . $generator->generateActiveField($attribute) . " ?>\n\n";
+<?php 
+$weights = [
+    'yii\validators\ExistValidator' => 2,
```
samdark: Could a simple ordered list instead of weights.

generators/crud/default/views/_form.php
```diff
@@ -28,15 +28,48 @@
...
+    'yii\validators\EmailValidator' => 0,
+    'yii\validators\NumberValidator' => 0,
+];
+
+$model = new $generator->modelClass();
+$attributes = array_flip($model->safeAttributes());
+
+foreach ($model->validators as $validator) {
+    if (!isset($weights[get_class($validator)])) {
+        continue;
```
samdark: It doesn't look right to me. Undefined validators should go to the end.

generators/crud/default/views/_form.php
```diff
@@ -28,15 +28,48 @@
...
+    }
+
+    $weight = $weights[get_class($validator)];
+    foreach ($validator->attributes as $attr) {
+        if (!is_array($attributes[$attr])
+            or $weight <= $attributes[$attr]['weight']
```
samdark: `or` → `||`

generators/crud/default/views/_form.php
```diff
@@ -28,15 +28,48 @@
...
+
+foreach ($attributes as $attr => $value) {
+    // code to generate each attribute based on the validators saved sofar.
+}
+    ?>
     <div class="form-group">
         <?= "<?= " ?>Html::submitButton($model->isNewRecord ? <?= $generator->generateString('Create') ?> : <?= $generator->generateString('Update') ?>, ['class' => $model->isNewRecord ? 'btn btn-success' : 'btn btn-primary']) ?>
     </div>
-
+
```
samdark: Remove spaces.

generators/crud/default/views/_form.php
Faryshta: at this point we are only finding out which validator has the biggest weight. The creation of the `$form->field($model, 'attribute')` part is bellow.

generators/crud/default/views/_form.php
Faryshta: what do you mean?

samdark:
```php
$priority = [
    'yii\validators\ExistValidator',
    'yii\validators\ImageValidator',
   // ...
];
```

Faryshta: then we should use something like

```php
$priority = array_flip([
    'yii\validators\ExistValidator',
    'yii\validators\ImageValidator',
   // ...
]);
```

ok will fix it

How about the line 66 which I think is the really problematic line because its whats missing to make this functional?

generators/crud/default/views/_form.php
```diff
@@ -28,15 +28,48 @@
...
+            continue;
+        }
+
+        $attributes[$attr] => ['weight' => $weight, 'validator' => $validator];
     }
-} ?>
+}
+
+foreach ($attributes as $attr => $value) {
+    // code to generate each attribute based on the validators saved sofar.
```
samdark:
```php
foreach ($weights as $validatorClass) {
    foreach ($attributes as $attr => $value) {
        // if class match, generate else skip
    }
}
```

Faryshta: you mean creating all the text inputs, then all the file inputs, then all the checkbox inputs and etc?

samdark: Ah, nope. Should be the other way around.

Faryshta: ok but the question is. How do we generate the `->input()` based on the validator used?

For example shall we use a giant `switch` with all the cases?

samdark: Probably yes. The switch could be moved into a method though.

Faryshta: ok i will write the switch and leave the architecture to you

Faryshta:
@samdark i made a commit with a mock up on how the switch will look like.

what do you think about it? and how do you wanna add it to a method in the generator class?

samdark:
I think it's OK. A method is a good idea.

Faryshta:
@samdark @cebe I think its ready to get tested http://pastie.org/10669273 here is an example of what I get with one of my most complex models

generators/crud/Generator.php
```diff
@@ -548,4 +550,133 @@ public function getColumnNames()
...
     }
+
+    public function validatorPriority(Validator $validator)
+    {
+        static $priorities = [
+            'yii\validators\NumberValidator' => 0,
+            'yii\validators\EmailValidator' => 1,
+            'yii\validators\BooleanValidator' => 2,
+            'yii\validators\RangeValidator' => 3,
+            'faryshta\\validators\\EnumValidator' => 4,
```
SilverFire: What does it do in the core? :)

Faryshta: i will take it out then. I was also thinking on a way to do it extendable so plugins could create their generators on gii without making an entire new gii.

any idea?

samdark: Make it non-static then and move it to module level.

samdark:
@Faryshta is it abandoned?

generators/crud/Generator.php
Faryshta: i don't understand this tip. can you explain further?

specially on how to let other modules or libraries define their own associated widgets for the validators they define such as my enum validator

MKiselev:
@samdark what need for accept PR? Refactoring?

generators/crud/Generator.php
samdark: This way it could be configured externally.

samdark: Also it doesn't really make sense to specify priorities as integers. Order of values in array should be enough.

generators/crud/Generator.php
```diff
@@ -548,4 +550,133 @@ public function getColumnNames()
...
+            '    ',
+            '        ',
+            '            ',
+            '                ',
+        ];
+
+        // code to generate each attribute based on the validators saved sofar.
+        switch ($validator->className()) {
+
+            case 'yii\validators\NumberValidator':
```
samdark: Instead of hardcoding we can introduce something like `NumberValidatorFieldGenerator`. This way introducing our own generators will be possible. These could be configured via what's called `$priorities` now.

samdark:
@MKiselev refactoring, tests, phpdoc, docs about how to expand this to your own validators.

Faryshta:
> how to expand this to your own validators.

with your suggestion of the static property i don't know how to do it myself

PR #79, opened 2015-08-23
Update form.php

phpniki:
I do not wish to have the schema name/database name added to my class name when generating relation classes. I believe this was the older functionality. I tracked down the fix needed to give the user the option and it appears when the "useSchemaName" option was added to the gii Generator somebody forgot to update the view in vendor/yiisoft/yii2-gii/generators/model/form.php to include the line.

echo $form->field($generator, 'useSchemaName')->checkbox();

Adding that one line in vendor/yiisoft/yii2-gii/generators/model/form.php will give the user the option to include or not the schema name in class generation. The label and hints were already coded.


samdark:
Thanks!

PR #85, opened 2015-09-01
CRUD code style fixes

arogachev:
- Corrected imports to be in alphabetical order.
- Removed extra line breaks and added new.
- When the number of params is 1, it should be "one liner" for more compact view (except very long line). If the user need to add more, he should add line breaks and extra comma on his own.
- Corrected PHPDoc (missing `@inheritdoc`, `more explicit `@return`, missing `@throws`).
- Prevented reaching max symbols limit in line.
- Encoding of title is redundant by default, because this string is safe and managed by programmer, not user.
- Removed extra concatenation with space.
- Converted double quotes to single used in empty string (`''` instead of `""`).
- Capitalized comments.
- Removed variable assignment in condition, I think it's bad practice and it's better to assign variable separately before, this makes code more clear and easier to read.

### DIFF
--- a/generators/crud/default/controller.php
+++ b/generators/crud/default/controller.php
@@ -29,11 +29,12 @@ echo "<?php\n";
 
 namespace <?= StringHelper::dirname(ltrim($generator->controllerClass, '\\')) ?>;
 
-use Yii;
 use <?= ltrim($generator->modelClass, '\\') ?>;
 <?php if (!empty($generator->searchModelClass)): ?>
-use <?= ltrim($generator->searchModelClass, '\\') . (isset($searchModelAlias) ? " as $searchModelAlias" : "") ?>;
+use <?= ltrim($generator->searchModelClass, '\\') . (isset($searchModelAlias) ? " as $searchModelAlias" : '') ?>;
+use Yii;
 <?php else: ?>
+use Yii;
 use yii\data\ActiveDataProvider;
 <?php endif; ?>
 use <?= ltrim($generator->baseControllerClass, '\\') ?>;
@@ -45,6 +46,9 @@ use yii\filters\VerbFilter;
  */
 class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->baseControllerClass) . "\n" ?>
 {
+    /**
+     * @inheritdoc
+     */
     public function behaviors()
     {
         return [
@@ -59,7 +63,7 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
 
     /**
      * Lists all <?= $modelClass ?> models.
-     * @return mixed
+     * @return string
      */
     public function actionIndex()
     {
@@ -72,32 +76,27 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
             'dataProvider' => $dataProvider,
         ]);
 <?php else: ?>
-        $dataProvider = new ActiveDataProvider([
-            'query' => <?= $modelClass ?>::find(),
-        ]);
+        $dataProvider = new ActiveDataProvider(['query' => <?= $modelClass ?>::find()]);
 
-        return $this->render('index', [
-            'dataProvider' => $dataProvider,
-        ]);
+        return $this->render('index', ['dataProvider' => $dataProvider]);
 <?php endif; ?>
     }
 
     /**
      * Displays a single <?= $modelClass ?> model.
      * <?= implode("\n     * ", $actionParamComments) . "\n" ?>
-     * @return mixed
+     * @return string
+     * @throws NotFoundHttpException if the model cannot be found
      */
     public function actionView(<?= $actionParams ?>)
     {
-        return $this->render('view', [
-            'model' => $this->findModel(<?= $actionParams ?>),
-        ]);
+        return $this->render('view', ['model' => $this->findModel(<?= $actionParams ?>)]);
     }
 
     /**
      * Creates a new <?= $modelClass ?> model.
      * If creation is successful, the browser will be redirected to the 'view' page.
-     * @return mixed
+     * @return string|\yii\web\Response
      */
     public function actionCreate()
     {
@@ -106,9 +105,7 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
         if ($model->load(Yii::$app->request->post()) && $model->save()) {
             return $this->redirect(['view', <?= $urlParams ?>]);
         } else {
-            return $this->render('create', [
-                'model' => $model,
-            ]);
+            return $this->render('create', ['model' => $model]);
         }
     }
 
@@ -116,7 +113,7 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
      * Updates an existing <?= $modelClass ?> model.
      * If update is successful, the browser will be redirected to the 'view' page.
      * <?= implode("\n     * ", $actionParamComments) . "\n" ?>
-     * @return mixed
+     * @return string|\yii\web\Response
      */
     public function actionUpdate(<?= $actionParams ?>)
     {
@@ -125,9 +122,7 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
         if ($model->load(Yii::$app->request->post()) && $model->save()) {
             return $this->redirect(['view', <?= $urlParams ?>]);
         } else {
-            return $this->render('update', [
-                'model' => $model,
-            ]);
+            return $this->render('update', ['model' => $model]);
         }
     }
 
@@ -135,7 +130,8 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
      * Deletes an existing <?= $modelClass ?> model.
      * If deletion is successful, the browser will be redirected to the 'index' page.
      * <?= implode("\n     * ", $actionParamComments) . "\n" ?>
-     * @return mixed
+     * @return \yii\web\Response
+     * @throws NotFoundHttpException if the model cannot be found
      */
     public function actionDelete(<?= $actionParams ?>)
     {
@@ -164,7 +160,8 @@ if (count($pks) === 1) {
     $condition = '[' . implode(', ', $condition) . ']';
 }
 ?>
-        if (($model = <?= $modelClass ?>::findOne(<?= $condition ?>)) !== null) {
+        $model = <?= $modelClass ?>::findOne(<?= $condition ?>);
+        if ($model !== null) {
             return $model;
         } else {
             throw new NotFoundHttpException('The requested page does not exist.');
--- a/generators/crud/default/search.php
+++ b/generators/crud/default/search.php
@@ -24,10 +24,10 @@ echo "<?php\n";
 
 namespace <?= StringHelper::dirname(ltrim($generator->searchModelClass, '\\')) ?>;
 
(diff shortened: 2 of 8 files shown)


generators/crud/default/views/create.php
```diff
@@ -19,12 +16,9 @@
...
 ?>
-<div class="<?= Inflector::camel2id(StringHelper::basename($generator->modelClass)) ?>-create">
 
-    <h1><?= "<?= " ?>Html::encode($this->title) ?></h1>
-
-    <?= "<?= " ?>$this->render('_form', [
-        'model' => $model,
-    ]) ?>
+<div class="<?= Inflector::camel2id(StringHelper::basename($generator->modelClass)) ?>-create">
+    <h1><?= "<?= " ?>$this->title</h1>
```
githubjeka: ууу куда дел Html::encode ?

generators/crud/default/controller.php
```diff
@@ -72,32 +76,27 @@ public function actionIndex()
             'dataProvider' => $dataProvider,
         ]);
 <?php else: ?>
-        $dataProvider = new ActiveDataProvider([
-            'query' => <?= $modelClass ?>::find(),
-        ]);
+        $dataProvider = new ActiveDataProvider(['query' => <?= $modelClass ?>::find()]);
```
samdark: It was OK as is.

generators/crud/default/controller.php
```diff
@@ -72,32 +76,27 @@ public function actionIndex()
...
 
-        return $this->render('index', [
-            'dataProvider' => $dataProvider,
-        ]);
+        return $this->render('index', ['dataProvider' => $dataProvider]);
```
samdark: It was OK as is.

generators/crud/default/controller.php
```diff
@@ -72,32 +76,27 @@ public function actionIndex()
...
-     * @return mixed
+     * @return string
+     * @throws NotFoundHttpException if the model cannot be found
      */
     public function actionView(<?= $actionParams ?>)
     {
-        return $this->render('view', [
-            'model' => $this->findModel(<?= $actionParams ?>),
-        ]);
+        return $this->render('view', ['model' => $this->findModel(<?= $actionParams ?>)]);
```
samdark: It was OK as is.

generators/crud/default/controller.php
```diff
@@ -72,32 +76,27 @@ public function actionIndex()
...
     }
 
     /**
      * Creates a new <?= $modelClass ?> model.
      * If creation is successful, the browser will be redirected to the 'view' page.
-     * @return mixed
+     * @return string|\yii\web\Response
```
samdark: In this case it's string.

generators/crud/default/controller.php
```diff
@@ -106,9 +105,7 @@ public function actionCreate()
         if ($model->load(Yii::$app->request->post()) && $model->save()) {
             return $this->redirect(['view', <?= $urlParams ?>]);
         } else {
-            return $this->render('create', [
-                'model' => $model,
-            ]);
+            return $this->render('create', ['model' => $model]);
```
samdark: It was OK as is.

generators/crud/default/controller.php
```diff
@@ -116,7 +113,7 @@ public function actionCreate()
      * Updates an existing <?= $modelClass ?> model.
      * If update is successful, the browser will be redirected to the 'view' page.
      * <?= implode("\n     * ", $actionParamComments) . "\n" ?>
-     * @return mixed
+     * @return string|\yii\web\Response
```
samdark: In this case it's string.

generators/crud/default/controller.php
```diff
@@ -125,9 +122,7 @@ public function actionUpdate(<?= $actionParams ?>)
         if ($model->load(Yii::$app->request->post()) && $model->save()) {
             return $this->redirect(['view', <?= $urlParams ?>]);
         } else {
-            return $this->render('update', [
-                'model' => $model,
-            ]);
+            return $this->render('update', ['model' => $model]);
```
samdark: It was OK as is.

generators/crud/default/controller.php
```diff
@@ -135,7 +130,8 @@ public function actionUpdate(<?= $actionParams ?>)
      * Deletes an existing <?= $modelClass ?> model.
      * If deletion is successful, the browser will be redirected to the 'index' page.
      * <?= implode("\n     * ", $actionParamComments) . "\n" ?>
-     * @return mixed
+     * @return \yii\web\Response
```
samdark: In this case it's string.

generators/crud/default/search.php
```diff
@@ -56,30 +56,26 @@ public function scenarios()
...
     {
         $query = <?= isset($modelAlias) ? $modelAlias : $modelClass ?>::find();
 
-        // add conditions that should always apply here
+        // Add conditions that should always apply here
 
-        $dataProvider = new ActiveDataProvider([
-            'query' => $query,
-        ]);
+        $dataProvider = new ActiveDataProvider(['query' => $query]);
```
samdark: It was OK as is.

generators/crud/default/views/create.php
samdark: Yep. `Html::encode` removal would lead to XSS.

generators/crud/default/views/create.php
```diff
@@ -19,12 +16,9 @@
...
 
+    <?= "<?= " ?>$this->render('_form', ['model' => $model]) ?>
```
samdark: It was OK as is.

generators/crud/default/views/index.php
```diff
@@ -22,23 +22,20 @@
 $this->title = <?= $generator->generateString(Inflector::pluralize(Inflector::camel2words(StringHelper::basename($generator->modelClass)))) ?>;
 $this->params['breadcrumbs'][] = $this->title;
 ?>
-<div class="<?= Inflector::camel2id(StringHelper::basename($generator->modelClass)) ?>-index">
 
-    <h1><?= "<?= " ?>Html::encode($this->title) ?></h1>
+<div class="<?= Inflector::camel2id(StringHelper::basename($generator->modelClass)) ?>-index">
+    <h1><?= "<?= " ?>$this->title</h1>
```
samdark: `Html::encode` again.

generators/crud/default/views/index.php
```diff
@@ -22,23 +22,20 @@
...
 <?php if(!empty($generator->searchModelClass)): ?>
 <?= "    <?php " . ($generator->indexWidgetType === 'grid' ? "// " : "") ?>echo $this->render('_search', ['model' => $searchModel]); ?>
 <?php endif; ?>
 
-    <p>
-        <?= "<?= " ?>Html::a(<?= $generator->generateString('Create ' . Inflector::camel2words(StringHelper::basename($generator->modelClass))) ?>, ['create'], ['class' => 'btn btn-success']) ?>
-    </p>
+    <p><?= "<?= " ?>Html::a(<?= $generator->generateString('Create ' . Inflector::camel2words(StringHelper::basename($generator->modelClass))) ?>, ['create'], ['class' => 'btn btn-success']) ?></p>
```
samdark: It was OK as is.

generators/crud/default/views/update.php
```diff
@@ -11,22 +11,17 @@
 echo "<?php\n";
 ?>
 
-use yii\helpers\Html;
-
 /* @var $this yii\web\View */
 /* @var $model <?= ltrim($generator->modelClass, '\\') ?> */
 
-$this->title = <?= $generator->generateString('Update {modelClass}: ', ['modelClass' => Inflector::camel2words(StringHelper::basename($generator->modelClass))]) ?> . ' ' . $model-><?= $generator->getNameAttribute() ?>;
+$this->title = <?= $generator->generateString('Update {modelClass}: ', ['modelClass' => Inflector::camel2words(StringHelper::basename($generator->modelClass))]) ?> . $model-><?= $generator->getNameAttribute() ?>;
```
samdark: Why removing space?

generators/crud/default/views/update.php
```diff
@@ -11,22 +11,17 @@
...
 ?>
-<div class="<?= Inflector::camel2id(StringHelper::basename($generator->modelClass)) ?>-update">
 
-    <h1><?= "<?= " ?>Html::encode($this->title) ?></h1>
-
-    <?= "<?= " ?>$this->render('_form', [
-        'model' => $model,
-    ]) ?>
+<div class="<?= Inflector::camel2id(StringHelper::basename($generator->modelClass)) ?>-update">
+    <h1><?= "<?= " ?>$this->title</h1>
```
samdark: `Html::encode`

generators/crud/default/views/view.php
```diff
@@ -21,9 +21,9 @@
 $this->params['breadcrumbs'][] = ['label' => <?= $generator->generateString(Inflector::pluralize(Inflector::camel2words(StringHelper::basename($generator->modelClass)))) ?>, 'url' => ['index']];
 $this->params['breadcrumbs'][] = $this->title;
 ?>
-<div class="<?= Inflector::camel2id(StringHelper::basename($generator->modelClass)) ?>-view">
 
-    <h1><?= "<?= " ?>Html::encode($this->title) ?></h1>
+<div class="<?= Inflector::camel2id(StringHelper::basename($generator->modelClass)) ?>-view">
+    <h1><?= "<?= " ?>$this->title</h1>
```
samdark: `Html::encode`

samdark:
> When the number of params is 1, it should be "one liner" for more compact view (except very long line). If the user need to add more, he should add line breaks and extra comma on his own.

It's more readable as is.

generators/crud/default/views/update.php
arogachev: What's the purpose of it?

samdark:
> Encoding of title is redundant by default, because this string is safe and managed by programmer, not user.

It won't hurt to encode in this case. Using model title as part of the title is very likely.

arogachev:
Disagree about "one liners" Why we need 3 lines instead of 1 when using just one parameter? For more than one, yes, we need separate it to different lines.

arogachev:
I don't understand about `Html::encode()`. Why we need encode strings such as `Create post`. What's unsafe in it? And it's managed only by programmer here. For more complex cases (for example when user input is involved), obviously, user need to add it, but on his own.

samdark:
> Why we need 3 lines instead of 1 when using just one parameter (and when this line is pretty short)?

From my experience, sooner or later, you're adding more arguments for `->render()` calls.

> Why we need encode strings such as Create post. What's unsafe in it?

Nothing until one decides to use post name in the edit form (which is common). Gii is a template so it's not a good idea to provide defaults which lead to insecure solutions.

arogachev:
About adding more arguments - isn't this premature optimization?

arogachev:
I understood you about `Html::encode()`, I think in this case we need to remove it only in `create` and `update` views, because there are no model titles. For `update` and `view ` views we can leave it as it initialliy was.

arogachev:
As for redirects - why it's string? It's more explicit variation than `@mixed`. Because there are only two options, I think it's better explicitly specify them.

klimov-paul:
Explicit specification will be `string|\yii\web\Response`.
As for me this is too long and makes controller code unecessary heavy.
Action methods usually are not invoked directly, so there is no benefit for marking exact result for them.

samdark:
Reviewed again. Won't accept it.

PR #86, opened 2015-09-17
added validate method on create and update action

leninhasda:
i have added the validate method for create and update action.
i am not sure whether it was not included for any reason or just forgotten, but i think this is an important one which should be there.

### DIFF
--- a/generators/crud/default/controller.php
+++ b/generators/crud/default/controller.php
@@ -103,7 +103,7 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
     {
         $model = new <?= $modelClass ?>();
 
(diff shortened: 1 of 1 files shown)


pana1990:
$model->save() use validation inside, this is no sense

leninhasda:
man, i really need to go through the doc in details. 
i followed a incorrect tuts on uploading file without a proper way to validate uploaded file which gave me a lot of trouble. and then i found validate function on form generator and thought it was some how left out.
thanks @pana1990 for the heads up :+1: 
now i know how it works and that i need to see the doc before doing anything stupid again :disappointed: 

i'm closing the pull

pana1990:
@leninhasda no problem men :)

PR #88, opened 2015-09-20
Generate inverse relation names

### LINKED ISSUE #54: Model generator, add inverseOf to relations.
I've just implemented this in my model generator. Filling those manually is very error prone (singular vs plural relation names, etc). This be included in core but I've noticed that not all relations need those (excluding many-many). I'd rather remove those unecessary than try to add the required ones.

Should there be an option to generate them, like _generateRelations_? Maybe _generateRelations_ should be a dropdown with options like:

* no relations
* generate relations
* generate relations with inverse

nineinchnick:
Resolves #54.

Expands the boolean _generateRelations_ option into a dropdown, adding a new option to generate inverse relation names.

### DIFF
--- a/generators/model/Generator.php
+++ b/generators/model/Generator.php
@@ -12,6 +12,7 @@ use yii\db\ActiveQuery;
 use yii\db\ActiveRecord;
 use yii\db\Connection;
 use yii\db\Schema;
+use yii\db\TableSchema;
 use yii\gii\CodeFile;
 use yii\helpers\Inflector;
 use yii\base\NotSupportedException;
@@ -24,12 +25,16 @@ use yii\base\NotSupportedException;
  */
 class Generator extends \yii\gii\Generator
 {
+    const RELATIONS_NONE = 'none';
+    const RELATIONS_ALL = 'all';
+    const RELATIONS_ALL_INVERSE = 'all-inverse';
+
     public $db = 'db';
     public $ns = 'app\models';
     public $tableName;
     public $modelClass;
     public $baseClass = 'yii\db\ActiveRecord';
-    public $generateRelations = true;
+    public $generateRelations = self::RELATIONS_ALL;
     public $generateLabelsFromComments = false;
     public $useTablePrefix = false;
     public $useSchemaName = true;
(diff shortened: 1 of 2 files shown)


generators/model/Generator.php
```diff
@@ -463,10 +466,76 @@ protected function generateRelations()
             }
         }
 
+        if ($this->generateRelations === self::RELATIONS_ALL_INVERSE) {
+            return $this->generateInverseRelations($relations);
```
samdark: Is it better to name it `addInverseRelations`?

generators/model/Generator.php
```diff
@@ -463,10 +466,76 @@ protected function generateRelations()
...
+    }
+
+    /**
+     * @param array $relations relation declarations
+     * @return array relation declarations extended with inverse relation names
+     */
+    protected function generateInverseRelations($relations)
+    {
+        $relationNames = [];
+        $db = $this->getDbConnection();
```
samdark: It's used only once. No need for a variable.

generators/model/Generator.php
```diff
@@ -463,10 +466,76 @@ protected function generateRelations()
...
+        }
         return $relations;
     }
 
     /**
+     * @param TableSchema $table
+     * @param array $fks
+     * @return bool
+     */
+    protected function isManyRelation($table, $fks)
```
samdark: Is it better to name it `isHasManyRelation`?

nineinchnick: I'm not sure. Having two verbs at the beginning is a bit confusing.

samdark: Well, relation type is `hasMany`, not `many`...

nineinchnick: It's stupid to argue about it, but I only found mentions of  [`one-to-many`](https://en.wikipedia.org/wiki/One-to-many_%28data_model%29) on Wikipedia :smile: 

Anyway, for consistency and whatever else needed to get this merged, I renamed it.

samdark: Yes. Classic ER types are `one-to-many`, `one-to-one`, `many-to-many` but that's not the case for Yii 2.0 ;)

samdark:
Merged https://github.com/yiisoft/yii2-gii/commit/a67035cd3225ea108db03a4da21b0204fe54630d. Thanks!

PR #90, opened 2015-09-28
remove useless function definitions

dizews:
I think it look like more clean then before

### DIFF
--- a/generators/model/default/query.php
+++ b/generators/model/default/query.php
@@ -22,6 +22,9 @@ namespace <?= $generator->queryNs ?>;
  * This is the ActiveQuery class for [[<?= $modelFullClassName ?>]].
  *
  * @see <?= $modelFullClassName . "\n" ?>
+ *
+ * @method <?= $modelFullClassName ?>[]|array all($db = null)
+ * @method <?= $modelFullClassName ?>|array|null one($db = null)
  */
 class <?= $className ?> extends <?= '\\' . ltrim($generator->queryBaseClass, '\\') . "\n" ?>
 {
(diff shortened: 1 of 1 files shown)


klimov-paul:
No, these definitions are placed for the IDE type-hinting.

dizews:
I think it was for type-hinting

dizews:
@klimov-paul do you want to say that my 
```php
+ *
+ * @method <?= $modelFullClassName ?>[]|array all($db = null)
+ * @method <?= $modelFullClassName ?>|array|null one($db = null)
```

would not work as IDE type-hinting? phpstorm work well.

samdark:
These should work. @dizews could you check NetBeans?

dizews:
@samdark it work in the NetBeans.

samdark:
@cebe is it a problem for our doc generator?

cebe:
it does not consider `@method` annotations yet so application docs will show docs for original implementation.

samdark:
OK. Then this pull request should wait till it's supported.

samdark:
@cebe is there any change in regards to doc tool?

cebe:
created an issue https://github.com/yiisoft/yii2-apidoc/issues/136

generators/model/default/query.php
```diff
@@ -31,21 +34,4 @@ class <?= $className ?> extends <?= '\\' . ltrim($generator->queryBaseClass, '\\
         return $this;
     }*/
```
arogachev, line 36: (Code style) Remove newline.

arogachev:
This can be merged now.

machour (approved):

samdark:
@arogachev there are conflicts to be resolved first.

PR #92, opened 2015-10-07
fix many-many relation names

nineinchnick:
Resolves yiisoft/yii2#9817. Use foreign key instead of primary key to generate relation name when using a junction table.


samdark:
Thanks!

samdark:
No changelog since it was recent stuff regression.

Issue #97, opened 2015-11-01
Convert backslashes to forward slashes in view paths


lubosdz:
Backslashes in view paths should be automatically converted to forward slash to avoid following common issues:
- @app\modules/admin/views/user -> will throw exception invalid alias (note backslash `\`)
- app/modules/admin/views/user  -> will create wrong dirs under `@web/app/modules/admin/views/user`

It is quite common to copy & paste namespaced paths from previous fields, or enter relative path (without @ alias prefix).

samdark:
The issue confirmed in:

- Controller generator view path.
- CRUD generator view path.
- Extension generator output path.

PR #98, opened 2015-11-01
#97 - Convert backslashes to forward slashes in view paths

lubosdz:
#97


samdark:
Does it apply to other generators? Also... your pull request seems to be messed up...

samdark:
This pull request can't be accepted as is.

Issue #102, opened 2015-11-11
'Check This File'-button in modal file-view doesn't update the 'Check All'-checkbox in the file list


Insensus:
### Issue
When previewing a file, the 'Check This File'-button at the top of the preview window doesn't update the state of the 'Check All'-checkbox above the file list.

### Cause
The function *checkFileToggle* in *gii.js:139* directly changes the checkbox-state instead of triggering a click event. Since no click event is fired, the state of the 'Check All'-checkbox does not get updated.

### Proposed solution
Trigger a click event on the checkbox instead of changing its state directly.

samdark:
Fixed with https://github.com/yiisoft/yii2-gii/commit/983a8b836c2097db553ace397358710ce89e1dfb. Thanks!

PR #103, opened 2015-11-11
'Check This File'-button in the file preview window triggers a click event on the checkbox instead of changing its state directly

Insensus:
#102


samdark:
Fixed with https://github.com/yiisoft/yii2-gii/commit/983a8b836c2097db553ace397358710ce89e1dfb. Thanks!

PR #129, opened 2016-02-02
Removed redundant return in active() scope example

arogachev:
We can just return `$this->andWhere`, [official docs reference](http://www.yiiframework.com/doc-2.0/guide-db-active-record.html#customizing-query-classes).


SilverFire:
Thank you!

PR #131, opened 2016-02-09
Use table comments for documentor's variable description

stmswitcher:
Hello!

I am using PHPdoc comments a lot in my work and in Yii projects I have to copy-paste the db fileds' descriptions into my models, most of which is generated by gii. So, I've made some changes in gii's model generator to use the db comments in model's doc-block.

It's not such a big thing, but may be usefull for those who is using PHPdoc in their IDE.

The changes include generator propery, corresponding checkbox and default model template.

### DIFF
--- a/generators/model/default/model.php
+++ b/generators/model/default/model.php
@@ -24,7 +24,7 @@ use Yii;
  * This is the model class for table "<?= $generator->generateTableName($tableName) ?>".
  *
 <?php foreach ($tableSchema->columns as $column): ?>
- * @property <?= "{$column->phpType} \${$column->name}\n" ?>
+ * @property <?= "{$column->phpType} \${$column->name}" . ($generator->generateDocsFromComments ? " {$column->comment}" : '') . "\n" ?>
 <?php endforeach; ?>
 <?php if (!empty($relations)): ?>
  *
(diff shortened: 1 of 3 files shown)


SilverFire:
Looks good to me

stmswitcher:
Btw, I don't use gii for enything but the models, so there's not any changes to other templates.

Faryshta:
what happens if my column comment has line breaks?

cebe:
@Faryshta good point, line breaks should be replaced by `"\n * "`.

generators/model/default/model.php
```diff
@@ -24,7 +24,7 @@
  * This is the model class for table "<?= $generator->generateTableName($tableName) ?>".
  *
 <?php foreach ($tableSchema->columns as $column): ?>
- * @property <?= "{$column->phpType} \${$column->name}\n" ?>
+ * @property <?= "{$column->phpType} \${$column->name}" . ($generator->generateDocsFromComments ? " {$column->comment}" : '') . "\n" ?>
```
Faryshta: `strtr($column->comment, ["\n" => "\n * "])`

michaelarnauts:
Is this feature considered to be included? Since this PR isn't up to date anymore, I'm willing to create a new one given that it can be merged.

Also, what about comments on the table itself? This information isn't currently available in the models, but it also could be read out (with a `SHOW CREATE TABLE xxx` in mysql for example). This would probably require a change in the db-classes.

cebe:
> Is this feature considered to be included?

yes, go ahead :)

PR #134, opened 2016-02-16
Attribute queryNs on model generator not validate.

zetamen:
On model generator validate only ns attribute. Fixed. Unit test added.


samdark:
Good catch. Verified it.

samdark:
Thank you.

PR #142, opened 2016-03-15
Removed redundant space in concatenation

arogachev:
Current generator generates something like:

```php
$this->title = 'Update Post: ' . ' ' . $model->id;
```

Space is already added after colon.


SilverFire:
Thank you!

PR #152, opened 2016-04-20
fix generating model without any rules

and800:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

Fix generating `return [,];` in model's `rules()` property when no rules have been created.


cebe:
could you add a changelog entry?

and800:
@cebe done, sorry for forgetting that

samdark:
Perfect. Thank you!

PR #153, opened 2016-04-24
Option to filter rows

thiagotalma:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/assets/gii.js
+++ b/assets/gii.js
@@ -157,9 +157,15 @@ yii.gii = (function ($) {
     };
 
     var initToggleActions = function () {
-        $('#action-toggle :input').change(function () {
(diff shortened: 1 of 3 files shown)


samdark:
Do we have that many files generated that a filter is required?

samdark:
Probably yes in case `*` is used. OK. Makes sense.

thiagotalma:
@samdark That's right. Do you get the idea.

Try to catch only the query files in a project with 57 tables that will understand what we are talking about.  :wink:

samdark:
Merged. Thanks!

PR #159, opened 2016-05-31
use `::class` constant for model generator

Faryshta:
| Q | A |
| --- | --- |
| Is bugfix? | no |
| New feature? | no |
| Breaks BC? | no |
| Tests pass? | yes |

### DIFF
--- a/generators/model/default/model.php
+++ b/generators/model/default/model.php
@@ -93,7 +93,7 @@ class <?= $className ?> extends <?= '\\' . ltrim($generator->baseClass, '\\') . 
      */
     public static function find()
     {
-        return new <?= $queryClassFullName ?>(get_called_class());
+        return new <?= $queryClassFullName ?>(static::class);
     }
 <?php endif; ?>
 }


generators/model/default/model.php
```diff
@@ -93,7 +93,7 @@ public function get<?= $name ?>()
      */
     public static function find()
     {
-        return new <?= $queryClassFullName ?>(get_called_class());
+        return new <?= $queryClassFullName ?>(static::class);
```
cebe: are you sure this works? `::class` is a compile time feature while `static` should work dynamically dependent on the call.

Faryshta: https://3v4l.org/0Yb1T

cebe: okay. still I find this a bit cryptic. get_called_class() is more obvious imo.

rob006: > but get get_called_class() is buggy. for example https://3v4l.org/LUafm

https://3v4l.org/Pdko5

klimov-paul:
Resolved by commit aa34d2174b3d1cc89fd9e80edbbc99460babea17

PR #162, opened 2016-06-14
support fk column with front id_

mootensai:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | my issues

Support foreign key column like id_user
not only user_id.


samdark:
Thanks!

PR #173, opened 2016-07-28
Changed label of attribute: Model Class Name

w3lifer:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no

It sound clearer than "Model Class" which associating with `app\models\Blablabla`. For example, in CRUD generator the "Model Class" mean fully qualified class name, and here this mean class name, not fully qualified.


samdark:
Thanks!

PR #184, opened 2016-08-26
Added PHPDoc to support the autocomplete for findOne() and findAll()

paulzi:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

### DIFF
--- a/generators/model/default/model.php
+++ b/generators/model/default/model.php
@@ -32,6 +32,9 @@ use Yii;
  * @property <?= $relation[1] . ($relation[2] ? '[]' : '') . ' $' . lcfirst($name) . "\n" ?>
 <?php endforeach; ?>
 <?php endif; ?>
+ *
+ * @method static <?= $className ?>|null findOne($condition)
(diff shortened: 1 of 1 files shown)


SilverFire:
Methods `findOne()` and `fineAll()` are declared in BaseActiveRecord and highlighting works without PHPDocs. What is the reason for this change? Or am I missing something?

paulzi:
findOne() and findAll() return self|null, in PHPStorm autocompletion does not work:
Sample

SilverFire:
Ah, than missing. @samdark do you know a better way?

samdark:
Does it work for `static` only w/o `null`?

paulzi:
Hmm.. Yes!
If I remove the `null` from PHPDoc `@return static|null` in the `BaseActiveRecord::findOne()` autocompletion works normally in PHPStorm. I'm close this pull-request and open new issue.

PR #219, opened 2016-11-16
Fixed a bulk model generation bug.

FROSTEROID:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no

It now _actually_ works as it is stated in the hint on table name field.
For example, table tbl_post will generate Post class, if the tableName field contains "tbl_*" pattern.
I'm actually surprised by the age of this bug. Nobody wanted to use that feature? :laughing:


samdark:
Checked with master code and I cannot reproduce the issue.

Created two tables:

- a_one
- a_two

Table name: `a_*`

Result: two models are generated as expected.

PR #226, opened 2016-12-08
Fill in class and property documentation based on Database Comments

michaelarnauts:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #131, 

Depends on https://github.com/yiisoft/yii2/pull/13162 for table comments.

I'm not sure if there has to be an option to enable this. This is just for comments, and doesn't have an impact on functionality or code.

### DIFF
--- a/generators/model/Generator.php
+++ b/generators/model/Generator.php
@@ -260,7 +260,11 @@ class Generator extends \yii\gii\Generator
             } else {
                 $type = $columnPhpType;
             }
-            $properties[$column->name] = $type;
+            $properties[$column->name] = [
+                'type' => $type,
+                'name' => $column->name,
+                'comment' => $column->comment,
+            ];
         }
 
         return $properties;
--- a/generators/model/default/model.php
+++ b/generators/model/default/model.php
@@ -22,10 +22,14 @@ namespace <?= $generator->ns ?>;
 use Yii;
 
 /**
+<?php if (property_exists($tableSchema, 'comment') && $tableSchema->comment): ?>
+ * <?= strtr($tableSchema->comment, ["\n" => "\n * "]) . "\n" ?>
+<?php else: ?>
  * This is the model class for table "<?= $generator->generateTableName($tableName) ?>".
+<?php endif; ?>
  *
-<?php foreach ($properties as $property => $type): ?>
- * @property <?= "{$type} \${$property}\n" ?>
+<?php foreach ($properties as $property => $data): ?>
+ * @property <?= "{$data['type']} \${$property}\n"  . ($data['comment'] ? ' ' . strtr($data['comment'], ["\n" => ' ']) : '') . "\n" ?>
 <?php endforeach; ?>
 <?php if (!empty($relations)): ?>
  *


generators/model/default/model.php
```diff
@@ -21,10 +21,14 @@
...
 /**
+<?php if (property_exists($tableSchema, 'comment') && $tableSchema->comment): ?>
+ * <?= strtr($tableSchema->comment, ["\n" => "\n * "]) . "\n" ?>
+<?php else: ?>
  * This is the model class for table "<?= $generator->generateTableName($tableName) ?>".
+<?php endif; ?>
  *
 <?php foreach ($tableSchema->columns as $column): ?>
- * @property <?= "{$column->phpType} \${$column->name}\n" ?>
+ * @property <?= "{$column->phpType} \${$column->name}" . ($column->comment ? " " . strtr($column->comment, ["\n" => "\n * "]) : '') . "\n" ?>
```
samdark: Can't be used like that since there's already an option in Gii Model generator "Generate Labels from DB Comments".

cebe: nope, labels are not generated from table comments, just column comments. these are not conflicting.

samdark: This line is about column comments.

cebe: right, but now I'm confused, how are column comments handled if not in this line!?

michaelarnauts: Why can't we use db comments for form labels and also property documentation at the same time? It does make sense, no?

cebe: okay, now I see. This line should respect the `$generateLabelsFromComments` and only add the comments here if they are not used as labels.

cebe: actually, you are right, @michaelarnauts we could still add them here, better have the label, than no documentation or description at all.

michaelarnauts:
What has kept this pull request from being merged? (It seems that is has gotten conflicts in the meantime). Can this be accepted when the conflicts are resolved?

michaelarnauts:
To recap, this doesn't change any code or changes the model in any way. It just makes sure that db comments are added to the phpdoc, so any IDE can use this to give additional information about the property or table object.

generators/model/Generator.php
```diff
@@ -260,7 +260,11 @@ private function generateProperties($table)
             } else {
                 $type = $columnPhpType;
             }
-            $properties[$column->name] = $type;
+            $properties[$column->name] = [
+                'type' => $type,
+                'name' => $column->name,
+                'description' => $column->comment,
+            ];
```
cebe: seems your last commit changed the diff to something wrong. This change looks unrelated to the changes below...

cebe: well, at least `comment` and `description` do not match.

michaelarnauts: I also need the comments, so it cant be just key => value.  The name is just for completeness.

cebe: sure, but you are creating the key `description` here, which you are using `comment` in the code below.

michaelarnauts: Ow, right. Sorry, will fix that.

cebe:
I have merged the property comment part, thank you! Keeping this open for the class comment when it is supported in the framework.

PR #227, opened 2016-12-19
Add default access control for controller methods

joshijitendra:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no

### DIFF
--- a/generators/crud/default/controller.php
+++ b/generators/crud/default/controller.php
@@ -39,6 +39,7 @@ use yii\data\ActiveDataProvider;
 use <?= ltrim($generator->baseControllerClass, '\\') ?>;
 use yii\web\NotFoundHttpException;
 use yii\filters\VerbFilter;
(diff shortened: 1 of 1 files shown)


samdark:
Does not make sense to me.  Would you please explain why the change is required?

joshijitendra:
This adds getting started code for access control filter which can improved further as required and saves time for having to add it manually for CRUD generator. Also it is in line with yii1 gii which has similar behaviour to add a default access control filter.

I am open to any better implementation/suggestion.

cebe:
I am for adding this because nearly all crud needs some access control and this way it is less likely to be forgotten. It could be configurable somehow. not sure how though.

samdark:
Thinking about it once again, it should be OK.

Sensetivity:
Maybe make sense do this feature optional?

Issue #238, opened 2017-01-27
Use int/bool instead of integer/boolean in doc blocks


klimov-paul:
### What steps will reproduce the problem?

Generate ActiveRecord for table containing interger or boolean fields

### What's expected?

Generated class PHPDocs displays attribute types in short name (PHP7 style). e.g. `@property int $id`

### What do you get instead?

Generated class PHPDocs displays attribute types in full name e.g. `@property integer $id`

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | *
| PHP version      | *
| Operating system | *

Back at Yii2 project there was a decision made to switch long names for `integer` and `boolean` types to the short ones: `int` and `bool` to match type names introduced by PHP 7.

See https://github.com/yiisoft/yii2/issues/12055

samdark:
Yes. Should be adjusted.

PR #239, opened 2017-01-28
Issue #238 fix

MKiselev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #238

### DIFF
--- a/generators/model/Generator.php
+++ b/generators/model/Generator.php
@@ -217,6 +217,7 @@ class Generator extends \yii\gii\Generator
                 'tableName' => $tableName,
                 'className' => $modelClassName,
                 'queryClassName' => $queryClassName,
+                'tableSchema' => $tableSchema,
                 'properties' => $this->generateProperties($tableSchema),
                 'labels' => $this->generateLabels($tableSchema),
                 'rules' => $this->generateRules($tableSchema),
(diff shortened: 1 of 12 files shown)


generators/model/Generator.php
```diff
@@ -242,6 +242,29 @@ public function generate()
     }
 
     /**
+     * Generates the properties for the specified table.
+     * @param \yii\db\TableSchema $table the table schema
+     * @return array the generated properties (property => type)
+     */
+    public function generateProperties($table)
```
samdark: Should it be public?

generators/model/Generator.php
```diff
@@ -242,6 +242,29 @@ public function generate()
     }
 
     /**
+     * Generates the properties for the specified table.
+     * @param \yii\db\TableSchema $table the table schema
+     * @return array the generated properties (property => type)
```
samdark: Need `@since`

generators/model/Generator.php
MKiselev: @samdark no, but methods e.g. generateLabels(), generateRules() also are public, but used only in current class.

samdark: Let's change visibility to private if it's not used anywhere. Or to protected if it could be overridden.

samdark:
Merged. Thank you!

PR #243, opened 2017-01-30
Issue #224 fix

MKiselev:
Add default validator with null value for integers when db is PostgreSQL

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #224


samdark:
Thanks!

PR #251, opened 2017-02-14
Fixes #249

### LINKED ISSUE #249: Gii should generate unique validator for tables with multiple primary keys
Table for example
```sql
CREATE TABLE "product_to_category" (
  product_id INTEGER NOT NULL,
  category_id INTEGER NOT NULL,
  PRIMARY KEY (product_id, category_id)
);
```
### What's expected?
Unique validator in rules section like that
```php
[['product_id', 'category_id'], 'unique', 'targetAttribute' => ['product_id', 'category_id'], 'message' => '']
```

### What do you get instead?
Nothing :smiley: 

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.11

**!Update**
I've not encountered this issue in sqlite, because sqlite does primary keys unique.

dmirogin:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #249

### DIFF
--- a/generators/model/Generator.php
+++ b/generators/model/Generator.php
@@ -360,6 +360,11 @@ class Generator extends \yii\gii\Generator
             foreach ($uniqueIndexes as $uniqueColumns) {
                 // Avoid validating auto incremental columns
                 if (!$this->isColumnAutoIncremental($table, $uniqueColumns)) {
+                    // Avoid validation for primary keys, we do it below
+                    if ($uniqueColumns === $table->primaryKey) {
+                        continue;
+                    }
+
                     $attributesCount = count($uniqueColumns);
 
                     if ($attributesCount === 1) {
@@ -391,6 +396,12 @@ class Generator extends \yii\gii\Generator
             }
             $targetAttributes = implode(', ', $targetAttributes);
             $rules[] = "[['$attributes'], 'exist', 'skipOnError' => true, 'targetClass' => $refClassName::className(), 'targetAttribute' => [$targetAttributes]]";
+        }
+
+        // Unique validation for multiple primary keys
+        if (is_array($table->primaryKey) && count($table->primaryKey) > 1) {
+            $attributes = implode("', '", $table->primaryKey);
+            $rules[] = "[['$attributes'], 'unique', 'targetAttribute' => ['$attributes']]";
         }
 
         return $rules;
--- a/generators/model/Generator.php
+++ b/generators/model/Generator.php
@@ -356,15 +356,11 @@ class Generator extends \yii\gii\Generator
 
         // Unique indexes rules
         try {
-            $uniqueIndexes = $db->getSchema()->findUniqueIndexes($table);
+            $uniqueIndexes = array_merge($db->getSchema()->findUniqueIndexes($table), [$table->primaryKey]);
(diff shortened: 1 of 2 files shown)


generators/model/Generator.php
```diff
@@ -393,6 +398,12 @@ public function generateRules($table)
             $rules[] = "[['$attributes'], 'exist', 'skipOnError' => true, 'targetClass' => $refClassName::className(), 'targetAttribute' => [$targetAttributes]]";
         }
 
+        // Unique validation for multiple primary keys
+        if (is_array($table->primaryKey) && count($table->primaryKey) > 1) {
+            $attributes = implode("', '", $table->primaryKey);
+            $rules[] = "[['$attributes'], 'unique', 'targetAttribute' => ['$attributes']]";
+        }
+
```
rob006: Why are you handle only composite PKs? This should be generated for all PKs which does not have autoincrement.

generators/model/Generator.php
```diff
@@ -360,6 +360,11 @@ public function generateRules($table)
             foreach ($uniqueIndexes as $uniqueColumns) {
                 // Avoid validating auto incremental columns
                 if (!$this->isColumnAutoIncremental($table, $uniqueColumns)) {
+                    // Avoid validation for primary keys, we do it below
+                    if ($uniqueColumns === $table->primaryKey) {
```
rob006: Are you sure this check is safe? The order of elements in the array is relevant.
https://3v4l.org/VYNjZ

dmirogin: For example sqlite create unique index for primary keys and it will the same with composite PK.
And i omit create validation in this block and handle all cases below.

generators/model/Generator.php
dmirogin: Okay, you're right, i'll do that.

generators/model/Generator.php
```diff
@@ -360,6 +360,11 @@ public function generateRules($table)
...
+                        continue;
+                    }
```
dynasource: you are postponing the logic here. This is error prone because what related is actually split which ideally should not be done with a comment.

generators/model/Generator.php
```diff
@@ -393,6 +398,12 @@ public function generateRules($table)
             $rules[] = "[['$attributes'], 'exist', 'skipOnError' => true, 'targetClass' => $refClassName::className(), 'targetAttribute' => [$targetAttributes]]";
         }
 
+        // Unique validation for multiple primary keys
+        if (is_array($table->primaryKey) && count($table->primaryKey) > 1) {
+            $attributes = implode("', '", $table->primaryKey);
+            $rules[] = "[['$attributes'], 'unique', 'targetAttribute' => ['$attributes']]";
+        }
```
dynasource: This solution is not practicing DRY with respect to https://github.com/dmirogin/yii2-gii/blob/df7bf1893f4a1aaf0de4e8255af1f04880f6ae16/generators/model/Generator.php#L374

samdark:
Merged in https://github.com/yiisoft/yii2-gii/commit/11dbba735105a098f8e0b44b601f91c062fb9324. Thanks!

PR #244, opened 2017-01-30
Issue #223 fix Use ilike operator for search model when db is PostgreSQL

MKiselev:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #223

PR #256, opened 2017-03-07
Code improvements for PR #244

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #223 

Improvements I mentioned in code review to PR #244 - https://github.com/yiisoft/yii2-gii/pull/244/files#r104593319.

### DIFF
--- a/generators/crud/Generator.php
+++ b/generators/crud/Generator.php
@@ -426,7 +426,7 @@ class Generator extends \yii\gii\Generator
                     break;
                 default:
                     $likeKeyword = $this->getClassDbDriverName() === 'pgsql' ? 'ilike' : 'like';
-                    $likeConditions[] = "->andFilterWhere(['$likeKeyword', '{$column}', \$this->{$column}])";                    
+                    $likeConditions[] = "->andFilterWhere(['{$likeKeyword}', '{$column}', \$this->{$column}])";                    
                     break;
             }
         }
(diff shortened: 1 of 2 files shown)


generators/crud/Generator.php
```diff
@@ -425,11 +425,8 @@ public function generateSearchConditions()
                     $hashConditions[] = "'{$column}' => \$this->{$column},";
                     break;
                 default:
-                    if ($this->getClassDbDriverName() === 'pgsql') {
-                        $likeConditions[] = "->andFilterWhere(['ilike', '{$column}', \$this->{$column}])";
-                    } else {
-                        $likeConditions[] = "->andFilterWhere(['like', '{$column}', \$this->{$column}])";
-                    }
```
cebe: in the above code it is easier to directly spot the difference between pgsql and others, so I am unsure about the change.

arogachev: @cebe Let me disagree with you. In current state even by a brief look these lines look similar, eyes are still scanning all of the symbols in both lines to detect possible differences other than `like` keywords. And even the amount of duplicated code is small, it's still copy-paste and violation of DRY. This is not case where we can violate DRY for readability purposes (for example, in tests) IMO.

So, even it's kind of microoptimization, I think it makes sense here.

cebe: thanks for disagreing with me. I'm fine with the changes :)

PR #259, opened 2017-03-14
Update composer.json

loveorigami:
For include bootstrap 3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     "minimum-stability": "dev",
     "require": {
         "yiisoft/yii2": "~2.0.4",
-        "yiisoft/yii2-bootstrap": "~2.0",
+        "yiisoft/yii2-bootstrap": "~2.0.0",
         "phpspec/php-diff": ">=1.0.2",
         "bower-asset/typeahead.js": "0.10.* | ~0.11.0"
     },


cebe:
will it break with bootstrap4 ?

samdark:
@loveorigami could you explain the change?

loveorigami:
@cebe 
today I clear cache in my composer and install packages. 
```"yiisoft/yii2-bootstrap": "~2.0"```
I got bootstrap 4

loveorigami:
Next I change to ```"yiisoft/yii2-bootstrap": "~2.0.0"```
And get only 2.0.0 version

samdark:
OK. Thanks for explaining. Merged.

cebe:
this looks like you messed something up with your composer.json,

1. "twbs/boostrap" should not be loaded in any case.
2. downgrade does not work well with composer asset plugin unless you disable this feature: https://github.com/fxpio/composer-asset-plugin/blob/master/Resources/doc/index.md#disable-the-import-filter-using-the-installed-packages

The problem does not seem related to this entry in composer.json.
If gii works without problems with bootstrap4, we should not limit the dependency here.

samdark:
Anyway, the change is consistent with project templates and overall it makes sense to limit version like that.

cebe:
I'd still prefer to check whether gii works with bootstrap4. Otherwise it would be impossible to use Gii 2.0.x with projects that use bootstrap4, and we'd need a 2.1 release of Gii only because of the bootstrap version constraint. Would be good to avoid that if possible.

PR #267, opened 2017-04-10
Fix an invalid type hint in a model template

Kolyunya:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | 

I may be missing something but it seems to me that `ActiveQueryInterface` would be more precise here since `\yii\db\BaseActiveRecord` [checks for it specifically](https://github.com/yiisoft/yii2/blob/master/framework/db/BaseActiveRecord.php#L286).

```php
    public function __get($name)
    {
        if (isset($this->_attributes[$name]) || array_key_exists($name, $this->_attributes)) {
            return $this->_attributes[$name];
        } elseif ($this->hasAttribute($name)) {
            return null;
        } else {
            if (isset($this->_related[$name]) || array_key_exists($name, $this->_related)) {
                return $this->_related[$name];
            }
            $value = parent::__get($name);
            if ($value instanceof ActiveQueryInterface) {
                return $this->_related[$name] = $value->findFor($name, $this);
            } else {
                return $value;
            }
        }
    }
```

**Related:** Am I right that these type hints([1](https://github.com/yiisoft/yii2/blob/master/framework/db/BaseActiveRecord.php#L410), [2](https://github.com/yiisoft/yii2/blob/master/framework/db/BaseActiveRecord.php#L369)) are invalid too and should read as `ActiveQuery`?

### DIFF
--- a/generators/model/default/model.php
+++ b/generators/model/default/model.php
@@ -76,7 +76,7 @@ class <?= $className ?> extends <?= '\\' . ltrim($generator->baseClass, '\\') . 
 <?php foreach ($relations as $name => $relation): ?>
 
     /**
-     * @return \yii\db\ActiveQuery
+     * @return \yii\db\ActiveQueryInterface
      */
     public function get<?= $name ?>()
     {


samdark:
> Related

Most probably yes.

samdark:
Merged. Thanks!

Kolyunya:
The problem was that type hints were not adjust. User code inside relation getter was returning `ActiveQueryInterface` (e.g. `$this->hasOne(...)`) while generated type hint was `ActiveQuery`.

Shouldn't we hasve fixed it the other was around?

samdark:
Ohh... that should be reverted :) I was too fast.

samdark:
Since Gii generates DB-backed models, it's fine to hint for more specific query.

Kolyunya:
But it seems that it can generate generic models.
```php
 class <?= $className ?> extends <?= '\\' . ltrim($generator->baseClass, '\\') . "\n" ?>
```

samdark:
By generic do you mean not backed by SQL database?

Kolyunya:
I thought that this template could be used in future for generating models other than `\yii\db\ActiveRecord`. Not sure if it makes sense.

Kolyunya:
Thinking of it again I'm not sure if we can fix it the other way around. `\yii\db\BaseActiveRecord::hasOne` returns `\yii\sphinx\ActiveQuery` for `\yii\sphinx\ActiveRecord` models. So we can't chage `hasOne` and `hasMany` hints.

Kolyunya:
We can either revert the revert and loosen the return type hint to `\yii\db\ActiveQueryInterface' or we can hint the return type in method body like this:
```
    /**
     * @return \yii\db\ActiveQuery
     */
    public function getFoo()
    {
        /* var $relation \yii\db\ActiveQuery */
        $relation = $this->hasOne(Foo::className(), ['id' => 'foo_id']);
        return $relation;
    }
```

samdark:
Why's that needed?

Kolyunya:
Because that's a type hint error. We declare two things:
* `getFoo` returns `\yii\db\ActiveQuery` 
* `$this->hasOne()` returns `\yii\db\ActiveQueryInterface`

`\yii\db\ActiveQueryInterface` is not an `\yii\db\ActiveQuery` and it's not compatible with it.

samdark:
Ah, right. Method body looks OK to me.

Kolyunya:
Should I make another PR?

samdark:
Yes if you don't mind doing it. Thanks.

Kolyunya:
Sure! Going to post it in a while.

cebe:
`yii\db\ActiveRecord::hasOne()` returns a `yii\db\ActiveQuery`, there is no reason to generalize that to the interface. https://github.com/yiisoft/yii2/blob/11761ac42a2e11fbc65a731675b270f16d37fef5/framework/db/ActiveRecord.php#L72-L73

samdark:
@cebe that's IDE inspection issue inside the method, not out of the method. Little thing but quite annoying.

Kolyunya:
@cebe thanks for the reference, I didn't know that hint existed. Seems like that is sufficient?

Ok, what about this?
```php
    public function hasOne($class, $link)
    {
        /* @var $class ActiveRecordInterface */
        /* @var $query ActiveQuery */
        $query = $class::find();
        $query->primaryModel = $this;
        $query->link = $link;
        $query->multiple = false;
        return $query;
    }
```

It seems to me that 
```php
        /* @var $query ActiveQuery */
        $query = $class::find();
```
is incorrect since it returns e.g. `yii\elasticsearch\ActiveQuery` not the `yii\db\ActiveQuery` for the `yii\elasticsearch\ActiveRecord`. Shouldn't we alter it?

cebe:
Not sure I fully understand the issue you are trying to solve here.

Kolyunya:
```php
        /* @var $query ActiveQuery */
        $query = $class::find();
```
This type hint seems to be invalid. It evaluates to the `\yii\db\ActiveQuery` whilst for `yii\sphinx\ActiveRecord` the actual type would be `\yii\sphinx\ActiveQuery`.

This seems to cause no problems but it's... invalid? If we're OK with that - let's leave it as it is.

cebe:
the most common case is creating code for db\ActiveRecord, so it should cover that case well. We could adjust the typehint in case a model is generated for sphinx, but using the interface instead does not make sense as the generator works only for SQL dbs.

Kolyunya:
```php
        /* @var $query ActiveQuery */
        $query = $class::find();
```
This is the code from the `yii\db\BaseActiveRecord`, not the template.

cebe:
okay, I am fine with changing that line.

cebe:
https://github.com/yiisoft/yii2/commit/0f78008bc71af6aca45152452f9f5017e7361e6a

Kolyunya:
```php
        /* @var $class ActiveRecordInterface */
        /* @var $query ActiveQuery */
        $query = $class::find();
```

`ActiveRecordInterface::find()` returns `ActiveQueryInterface`. The second hint is excessive, isn't it?

cebe:
hm yeah, but now I see why it hints ActiveQuery instead of ActiveQueryInterface. It accesses the properties of ActiveQuery, which are obviously not present on the interface...

Kolyunya:
Yeah, that's the problem too.

cebe:
I'd keep everything as is then.

PR #271, opened 2017-04-27
Update action.php

amin0256:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | no
| Fixed issues  | none


samdark:
Why?

amin0256:
@samdark because when you copy and paste that in a controller, that not work.
controller is in `@app/controllers/TestController` and model is in `@app/models/Test`, now compiler want to find it in `@app/controllers/app/models/Test`.
sorry for my bad language. 
tnx <3

cebe:
` $generator->modelClass ` should already be an absolute class name including the namespace

amin0256:
@cebe it will make a namespace like 'app\models\User' 
do you can check it ?

cebe:
The issue is valid but it should be ensured in the Generator that modelClass is absolute in all places.

cebe:
Merged, thank you!

amin0256:
thank you dear @cebe <3

Issue #287, opened 2017-07-28
Hint with target ActiveQuery class, not base ActiveQuery


bscheshirwork:
### What steps will reproduce the problem?
1.Use model generator 
2.Have relation
3.Produce ActiveQuery

### What's expected?
relation docs depend on target  ActiveQuery class
```
    /**
     * @return AnotherModelQuery
     */
    public function getAnotherModels()
    {
        return $this->hasMany(AnotherModel::className(), ['anotherId' => 'id'])->inverseOf('firstModel');
    }
```

### What do you get instead? 
```
    /**
     * @return \yii\db\ActiveQuery
     */
    public function getAnotherModels()
    {
        return $this->hasMany(AnotherModel::className(), ['anotherId' => 'id'])->inverseOf('firstModel');
    }
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      |dev
| PHP version      |7.1.7
| Operating system |

samdark:
Would be nice to have it indeed.

bscheshirwork:
I think if one time use checkbox "generate ActiveQuery" - any time use it. So... 
```
     * @return <?= ($queryClassName ? $relation[1] . 'Query' : '\yii\db\ActiveQuery') . "\n" ?>
```
another case - with asterisks - same result

klimov-paul:
No this won't work as actual query class name may not follow such convention.
There should be call of `ActiveRecord::find()` method to determine actual class name.

klimov-paul:
> But for all cases may be need |?

No it is redundant.

bscheshirwork:
upd with `ActiveRecord::find()`

bscheshirwork:
```
    /**
     * @return <?php
        if ($queryClassName) {
            $queryClassRealName = '\\' . $generator->queryNs . '\\' . $relation[1];
            if (class_exists($queryClassRealName, true) && is_subclass_of($queryClassRealName, '\yii\db\BaseActiveRecord')) {
                /** @var \yii\db\ActiveQuery $activeQuery */
                $activeQuery = $queryClassRealName::find();
                $activeQueryClass = $activeQuery::className();
                if(strpos($activeQueryClass, $generator->ns) === 0){
                    $activeQueryClass = StringHelper::basename($activeQueryClass);
                }
                echo $activeQueryClass;
            } else {
[...]

PR #288, opened 2017-07-28
Hint with target ActiveQuery class

bscheshirwork:
```
    /**
     * @return AnotherModelQuery
     */
    public function getAnotherModels()
    {
        return $this->hasMany(AnotherModel::className(), ['anotherId' => 'id'])->inverseOf('firstModel');
    }
```

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes/no
| Tests pass?   | yes/no
| Fixed issues  | #287


bscheshirwork:
up #287

generators/model/default/model.php
```diff
@@ -76,7 +76,25 @@ public function attributeLabels()
 <?php foreach ($relations as $name => $relation): ?>
 
     /**
-     * @return \yii\db\ActiveQuery
+     * @return <?php
+        if ($queryClassName) {
```
samdark: How about extracting it into model method?

generators/model/Generator.php
```diff
@@ -296,6 +299,37 @@ public function generateLabels($table)
...
+            if ($generateQuery) {
+                $queryClassRealName = '\\' . $this->queryNs . '\\' . $relation[1];
+                if (class_exists($queryClassRealName, true) && is_subclass_of($queryClassRealName, '\yii\db\BaseActiveRecord')) {
+                    /** @var \yii\db\ActiveQuery $activeQuery */
+                    $activeQuery = $queryClassRealName::find();
+                    $activeQueryClass = $activeQuery::className();
+                    if(strpos($activeQueryClass, $this->ns) === 0){
+                        $activeQueryClass = StringHelper::basename($activeQueryClass);
+                    }
+                    $result[$name] = '\yii\db\ActiveQuery|' . $activeQueryClass;
```
samdark: Why do we need `\yii\db\ActiveQuery|` part here? Isn't `$activeQueryClass` inherited from it already?

generators/model/Generator.php
```diff
@@ -296,6 +299,37 @@ public function generateLabels($table)
...
+            if ($generateQuery) {
+                $queryClassRealName = '\\' . $this->queryNs . '\\' . $relation[1];
+                if (class_exists($queryClassRealName, true) && is_subclass_of($queryClassRealName, '\yii\db\BaseActiveRecord')) {
+                    /** @var \yii\db\ActiveQuery $activeQuery */
+                    $activeQuery = $queryClassRealName::find();
+                    $activeQueryClass = $activeQuery::className();
+                    if (strpos($activeQueryClass, $this->ns) === 0){
+                        $activeQueryClass = StringHelper::basename($activeQueryClass);
+                    }
+                    $result[$name] = '\yii\db\ActiveQuery|' . $activeQueryClass;
```
samdark: Why do we need `\yii\db\ActiveQuery|` part here? Isn't `$activeQueryClass` inherited from it already?

bscheshirwork: Mismatch of return type:

`hasMany` returns `\yii\db\ActiveQuery`, not target $activeQueryClass. And another build-in functional don't know about children classes.
Redefine all build-in function each time is not a way.

samdark: Huh? Isn't inheritance properly handled by IDEs even if we don't override all methods of the class?

bscheshirwork: I think we don't have a tool for a define child class in PHPDoc of parent class or as return type of parent class 
https://github.com/yiisoft/yii2/blob/9b01ca275f891c245c2e840ed93c87f715c04e70/framework/db/ActiveRecord.php#L72-L73
```php
 * @method ActiveQuery hasMany($class, array $link) see [[BaseActiveRecord::hasMany()]] for more info
 * @method ActiveQuery hasOne($class, array $link) see [[BaseActiveRecord::hasOne()]] for more info
```
not accept `static` or some else like `$this`. I right?.. NO!

Hmmm... I find a solution! :smile_cat:

```php
 * @method $this hasMany($class, array $link) see [[BaseActiveRecord::hasMany()]] for more info
 * @method $this hasOne($class, array $link) see [[BaseActiveRecord::hasOne()]] for more info
```

What about it?

samdark: Looks good.

bscheshirwork: So... This is some harder.

I can propose new interface `RelatedActveQueryInterface` and set it in implementation in ActiveRecord

or leave template as is

bscheshirwork:
https://github.com/yiisoft/yii2/issues/14760

generators/model/Generator.php
bscheshirwork: Nope not yet...

This is solution for same class, not another

bscheshirwork:
@samdark no way to avoid "|" :crying_cat_face:

> there is no way to solve this in Yii itself.

samdark:
@bscheshirwork do you plan working more on this one?

bscheshirwork:
"working more"
Anything else needed?
p.s.:merge actual master

src/generators/model/default/model.php
```diff
@@ -76,7 +76,7 @@ public function attributeLabels()
 <?php foreach ($relations as $name => $relation): ?>
 
     /**
-     * @return \yii\db\ActiveQuery
+     * @return <?= $relationsClassHints[$name] . "\n" ?>
```
rustamwin: new line needed?

samdark: Nope.

samdark:
Merged. Thank you!

PR #293, opened 2017-08-31
Simplify templates

bscheshirwork:
Don't need `else` after `return`
Like `codeclimate` issues

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | will change result of generator work


samdark:
Merged. Thanks!

PR #296, opened 2017-09-02
Support fk column with code_ or _code

adipriyantobpn:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | my issues

Add support for foreign key column like `code_supplier` or `supplier_code`

### DIFF
--- a/generators/model/Generator.php
+++ b/generators/model/Generator.php
@@ -684,11 +684,17 @@ class Generator extends \yii\gii\Generator
             $baseModel = new $baseClass();
             $baseModel->setAttributes([]);
         }
-        if (!empty($key) && strcasecmp($key, 'id')) {
-            if (substr_compare($key, 'id', -2, 2, true) === 0) {
-                $key = rtrim(substr($key, 0, -2), '_');
-            } elseif (substr_compare($key, 'id', 0, 2, true) === 0) {
-                $key = ltrim(substr($key, 2, strlen($key)), '_');
+        $string_replacements = [
+            'id',
+            'code',
+        ];
+        foreach ($string_replacements as $str) {
+            if (!empty($key) && strcasecmp($key, $str)) {
+                if (substr_compare($key, $str, strlen($str)*-1, strlen($str), true) === 0) {
+                    $key = rtrim(substr($key, 0, strlen($str)*-1), '_');
+                } elseif (substr_compare($key, $str, 0, strlen($str), true) === 0) {
+                    $key = ltrim(substr($key, strlen($str), strlen($key)), '_');
+                }
             }
         }
         if ($multiple) {


adipriyantobpn:
AFAIK, it is also common to use `code` as prefix or suffix in foreign key column.

In my case, actually we are need to build REST API for a legacy desktop application.
As that application is developed by vendor, we only have authorization to access the database, but not change the database structure (as if changing the foreign key column from `supplier_code` into `supplier_id`).

It will be nice, if string `code` also can be cleaned-up in relation name generation, in addition to string `id`.
As in my case, the database contains almost 300 tables, which always have `code` prefix or suffix in its foreign key columns.

> When does the issue occur?

August 2017, since at that time we'd just start the REST API development.

> What do you see? And what was the expected result?

All relation name which it's foreign key has `code` prefix or suffix, is generated with `Code` in it.
For example for foreign key `supplier_code`, instead of getting `getSupplier()` for its relation function name, gii generate `getSupplierCode()`

samdark:
OK. Got it. I can't say that using `code` for the purpose is common.

Instead of introducing hardcoded `$string_replacements` it worth making a list of replacements  a property in model generator and leaving its default value to `['id']`. This way there will be no false positives and you'll be able to set your own custom replacements via gii config.

adipriyantobpn:
@samdark 
Thank you so much for your feedback.
Noted. I'll make public property for that.

dmirogin:
Please, write unit tests for this case.

generators/model/Generator.php
```diff
@@ -684,11 +685,13 @@ protected function generateRelationName($relations, $table, $key, $multiple)
             $baseModel = new $baseClass();
             $baseModel->setAttributes([]);
         }
-        if (!empty($key) && strcasecmp($key, 'id')) {
-            if (substr_compare($key, 'id', -2, 2, true) === 0) {
-                $key = rtrim(substr($key, 0, -2), '_');
-            } elseif (substr_compare($key, 'id', 0, 2, true) === 0) {
-                $key = ltrim(substr($key, 2, strlen($key)), '_');
+        foreach ($this->fkColumnIdentifiers as $str) {
+            if (!empty($key) && strcasecmp($key, $str)) {
```
samdark: What's meant by `strcasecmp` here? Not equal?

adipriyantobpn: Yes, i think so.
Actually i just copy that line from [`if (!empty($key) && strcasecmp($key, 'id')) {`](https://github.com/yiisoft/yii2-gii/blob/4ec2b55e4f0e5a6f27be93b4a1618a727409dec5/generators/model/Generator.php#L687) and change `id` string into `$str` variable accordingly.

Is there anything that i should revise?

adipriyantobpn:
@dmirogin, as a matter of fact, at this time, i don't have much experience for writing unit testing yet.
But i would learn it for sure.

NickvdMeij:
@samdark these changes _might_ have impact on https://github.com/yiisoft/yii2-gii/pull/417 . was looking through the milestone issues and noticed this. Haven't checked it myself to be sure

PR #301, opened 2017-09-15
Feature/2.1 no assets

schmunk42:
Adjustments for 2.1. Wanted to merge against `2.1` branch, but it's way more behind.

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | ?

Detailed description will follow in https://github.com/yiisoft/yii2/issues/8452


composer.json
```diff
@@ -19,10 +19,9 @@
     ],
     "minimum-stability": "dev",
     "require": {
-        "yiisoft/yii2": "~2.0.4",
-        "yiisoft/yii2-bootstrap": "~2.0.0",
-        "phpspec/php-diff": ">=1.0.2",
-        "bower-asset/typeahead.js": "0.10.* | ~0.11.0"
```
klimov-paul: Simple removal of `typeahead.js` is unacceptable - it will break the web-interface.
Unless there is some check and walkaround for its absence - it could not be simply removed.

klimov-paul:
Actually I can not see how this PR solves 'no asset' task, at least in its present state.
Also I would like do not mix later PHP version features apply with assets dealing in the single changeset.
Patch for `Model::class` and other PHP > 5.4 constructions is better to be a sepated PR.

klimov-paul:
For the Gii solving 'no assets' feature will be providing an alternative web-interface (may be theme or something), which will not use JS (or at least JQuery) but use old-fasion page-refreshing approach.
Such interface will be less pretty, but it will function.

Depending on particular solutions for the client script handling at `yii\web\ActiveForm` solution may vary from theme to particular widget's behavior class.

Saying that developer should write `packages.json` and install dependencies just to run the code generator does not feel right,

composer.json
schmunk42: `typeahead.js` (if required) should be moved to a `packages.json` file in this repo (or removed, if the code is rewritten). If it's moved to `packages.json` it can be installed with `foxypkg`, see [example](https://github.com/schmunk42/foxy-roxy-rabbit/blob/bfbacd9d01d4aacc79819d749cfe5e1fe768f9b5/pkg/package.json#L8)

schmunk42:
> Patch for Model::class and other PHP > 5.4 constructions is better to be a sepated PR.

I agree, will change this after the other issues have been discussed.

schmunk42:
> Saying that developer should write packages.json and install dependencies just to run the code generator does not feel right.

This is a general thing, if there are extensions which require client-side packages, how would they have to be installed?

klimov-paul:
> This is a general thing, if there are extensions which require client-side packages, how would they have to be installed?

For `yii2-gii` and `yii2-debug` it is not a general thing. These extensions are the development tools, which function should be garanteed. They should be available to function without extra JS dependencies installed.
Keep in mind that `yii2-gii` has a console interface, which is requirement free already. There also should be a view theme for the web interface, which is JQuery independent.

schmunk42:
> They should be available to function without extra JS dependencies installed.

I'd be totally fine with this. But what's about the bootstrap CSS dependency?

> Keep in mind that yii2-gii has a console interface, which is requirement free already.

Sure, I am almost using the console tools only.

PS: I've create two separate PRs for this
- https://github.com/yiisoft/yii2-gii/pull/302
- https://github.com/yiisoft/yii2-gii/pull/303

PR #317, opened 2017-11-26
Force HTML content type in response to display HTML when app is configured for REST API

microThread:
For times when developing a REST app and we want to access gii module from browser, the problem at this time is that the response format will be  yii\web\Response::FORMAT_JSON

| Q             | A
| ------------- | ---
| Is bugfix?   | yes
| New feature?  |no
| Breaks BC?    |no
| Tests pass?   |?

### DIFF
--- a/controllers/DefaultController.php
+++ b/controllers/DefaultController.php
@@ -10,6 +10,7 @@ namespace yii\gii\controllers;
 use Yii;
 use yii\web\Controller;
 use yii\web\NotFoundHttpException;
+use yii\web\Response;
 
 /**
  * @author Qiang Xue <qiang.xue@gmail.com>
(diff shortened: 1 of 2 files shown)


controllers/DefaultController.php
```diff
@@ -26,6 +27,13 @@ class DefaultController extends Controller
      * @var \yii\gii\Generator
      */
     public $generator;
+    
+    
+    public function beforeAction($action)
```
samdark:
```
/**
  * @inheritdoc
  */
```

microThread:
Done, I just added a new line into the CHANGELOG.md

samdark:
Merged. Thanks!

PR #320, opened 2017-12-02
Close #318 | Exception in Module generator with php7.2

### LINKED ISSUE #318: Exception in Module generator with php7.2
When I try to generate a new module in gii (using php7.2), it throws Exception 
### What's expected?
New module
### What do you get instead?
Exception Cannot use yii\base\Object as Object because 'Object' is a special class name

### How to fix it
In file /yiisoft/yii2-gii/CodeFile.php
1) `use yii\base\Object;` to `use yii\base\BaseObject;`
2) `class CodeFile extends Object` ---> `class CodeFile extends BaseObject`

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13.1
| PHP version      | 7.2
| Operating system | macOS

MKiselev:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #318

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Yii Framework 2 gii extension Change Log
 
 2.0.6 under development
 -----------------------
+- Bug #318: Use `yii\base\BaseObject` instead `yii\base\Object` in `CodeFile.php` (MKiselev)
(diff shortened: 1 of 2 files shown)


samdark:
Merged. Thank you!

Issue #333, opened 2018-02-16
Incorrect validation rule for TINYINT column type


nostop8:
### What steps will reproduce the problem?
1. Create table which have `tinyint` column, e.g. `status`
2. Goto Gii -> Models and generate model for your table
3. In preview see that `status` column gets inside `string` validation rules
### What's expected?
It should be inside `integer` validation rules as it always was.
### What do you get instead?
As it was mentioned, `status` column gets inside `string` validation rules
### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | dev-master 4518921
| PHP version      | 7.0
| Operating system | Windows

PR #334, opened 2018-02-16
Fixed incorrect validation rule for TINYINT column type

nostop8:
After recent updates to Yii2 Schema, Gii Model started to generate `string` validator rules for `tinyint` column type, where instead it should be `integer` validator as it always was. After these updates my applications started to show validation errors (particularly REST applications, did not test with client-server application).

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #333

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 gii extension Change Log
 2.0.7 under development
 -----------------------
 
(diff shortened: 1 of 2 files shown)


nostop8:
@samdark added line to changelog

samdark:
Merged. Thank you!

PR #344, opened 2018-03-07
fix 185 - Error when invalid FKs referencing non-existent table

adipriyantobpn:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Fixed issues  | #185 

Do same steps done by @samdark at commit aeec43d, but this time in `addInverseRelations()` function

### DIFF
--- a/src/Generator.php
+++ b/src/Generator.php
@@ -425,7 +425,8 @@ abstract class Generator extends Model
             'declare',
             'default',
             'die',
-            'do',
+            // COMPLETED_TODO : un-list do from reserved words, and map it to delivery order
+            // 'do',
             'echo',
             'else',
             'elseif',
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -562,6 +562,10 @@ class Generator extends \yii\gii\Generator
                 foreach ($table->foreignKeys as $refs) {
                     $refTable = $refs[0];
                     $refTableSchema = $db->getTableSchema($refTable);
+                    if ($refTableSchema === null) {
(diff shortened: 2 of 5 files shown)


src/Generator.php
```diff
@@ -425,7 +425,8 @@ public function isReservedKeyword($value)
             'declare',
             'default',
             'die',
-            'do',
+            // COMPLETED_TODO : un-list do from reserved words, and map it to delivery order
```
samdark: ?

src/generators/model/Generator.php
```diff
@@ -78,12 +86,14 @@ public function rules()
             [['ns', 'queryNs'], 'validateNamespace'],
             [['tableName'], 'validateTableName'],
             [['modelClass'], 'validateModelClass', 'skipOnEmpty' => false],
-            [['baseClass'], 'validateClass', 'params' => ['extends' => ActiveRecord::className()]],
-            [['queryBaseClass'], 'validateClass', 'params' => ['extends' => ActiveQuery::className()]],
+            [['baseClass'], 'validateClass', 'params' => ['extends' => ActiveRecord::class]],
```
samdark: Should not be changed. Next release if for Yii 2.0.

src/generators/model/Generator.php
```diff
@@ -43,6 +43,14 @@ class Generator extends \yii\gii\Generator
     public $queryNs = 'app\models';
     public $queryClass;
     public $queryBaseClass = 'yii\db\ActiveQuery';
+    // @TODO - extended model/query generation SHOULD BE CONFIGURABLE via boolean property
```
samdark: ?

Issue #327, opened 2018-01-02
Cannot instantiate abstract class common\models\AbstractActiveRecord


achretien:
### What steps will reproduce the problem?
Use an abstract class as $baseClass

### What's expected?
Generate the files properly.

### What do you get instead?
The error "Cannot instantiate abstract class common\models\AbstractActiveRecord" in `generators/model/Generator.php` at line 686.

https://github.com/yiisoft/yii2-gii/blob/f15b8dce4ef577c250112cf06b4616cbb2aca4ea/generators/model/Generator.php#L686

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13.1
| PHP version      | 7.1
| Operating system | alpine-linux

NickvdMeij:
Same problem here after updating to dev-master from v2.0.5

Yii version: 2.0.15.1
PHP version: 7.0.26
Operation System: Debian

PR #365, opened 2018-05-23
Fixed Model Generator to support abstract baseClass

rhertogh:
Checks if baseClass is an abstract class, and if so wrap it in a regular class before instantiating it.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #327 "Cannot instantiate abstract class common\models\AbstractActiveRecord"


rhertogh:
@samdark fixed

rhertogh:
I've removed the redundant class_exists check from the original pull request since ReflectionClass already checks if class exists.

samdark:
Merged. Thank you!

PR #369, opened 2018-06-12
Fix int instead of integer in phpdoc

tarasio:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | none


samdark:
It is correct as is.

tarasio:
Considering https://github.com/yiisoft/yii2-gii/issues/238#issuecomment-275671682 and [yii2 example](https://github.com/yiisoft/yii2/blob/master/framework/db/ActiveRecord.php#L738) you say it should be `@param integer` tag in phpdoc?

samdark:
It should be `int` and `bool`.

Issue #366, opened 2018-05-24
Wrong capitals for class/files when table is uppercase


slinstj:
### What steps will reproduce the problem?
yiisoft/yii2-gii uses `BaseInflector::id2camel()` to generate class names but, when getting upper case table names from db like `MY_TABLE`, id2camel() does not work properly.

If it is ok, I could do a PR removing id2camel() and doing an inline conversion inside `Generator::generateClassNames()`.

### What is the expected result?
```
'MY_TABLE' --> MyTable
```

### What do you get instead?
```
'MY_TABLE' --> MYTABLE
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.15.1
| PHP version      |  7.1.17
| Operating system | Win10

slinstj:
Related: https://github.com/yiisoft/yii2/issues/16316

samdark:
How would it work with `MyTable`?

slinstj:
Note sure what you mean but I would just change the generated class and file names. The table set in getTableName would not change. I have this working for MSSQL and MySQL (I extended the Generator).

On the other hand, I must to say that I am not familiar with Postgre.

slinstj:
AFAICS, it would not change. The change would affect only `MY_TABLE` since we have all uppercase and **underscores** separeting the words. `MyTable` would not be affected.

samdark:
How would you implement that?

slinstj:
```php
// I changed this in generateClassnames()
return $this->classNames[$fullTableName] = Inflector::id2camel($schemaName.$className, '_');

// For this (not beauty but works for my companie DBs [mysql, mssql])
return $this->classNames[$fullTableName] = strtr(ucwords(implode(' ', explode('_', strtolower($schemaName.$className)))), [' ' => '']);

// To implement it for MyTable I just would user Inflector::camel2words()
// (preferably in a bit more clean way than this)
return $this->classNames[$fullTableName] = strtr(ucwords(implode(' ', explode('_', strtolower(strtr(Inflector::camel2words($schemaName.$className), [' ' => '_']))))), [' ' => '']);

```

samdark:
It won't work: https://3v4l.org/164si

slinstj:
May be your test function `yiiCamelize()` is wrong.
Correct should be like `Inflector::camel2words()`:

```php
public static function camel2words($name, $ucwords = true)
    {
        $label = strtolower(trim(str_replace([
            '-',
            '_',
[...]

PR #370, opened 2018-06-13
366 Case insensitive class names for model generator

slinstj:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes

#366

### DIFF
--- a/tests/generators/ModelGeneratorTest.php
+++ b/tests/generators/ModelGeneratorTest.php
@@ -218,7 +218,7 @@ class ModelGeneratorTest extends GiiTestCase
     public function testGenerateClassNameForcingPsr()
     {
         $modelGenerator = new ModelGeneratorMock;
-        $modelGenerator->forcePsrClassNames = true;
+        $modelGenerator->caseInsensitive = true;
 
         $tableNames = [
             'lower_underline_name' => 'LowerUnderlineName',
@@ -242,7 +242,7 @@ class ModelGeneratorTest extends GiiTestCase
     public function testGenerateClassNameNoForcingPsr()
     {
         $modelGenerator = new ModelGeneratorMock;
-        $modelGenerator->forcePsrClassNames = false;
+        $modelGenerator->caseInsensitive = false;
 
         $tableNames = [
             'lower_underline_name' => 'LowerUnderlineName',
--- a/tests/generators/ModelGeneratorTest.php
+++ b/tests/generators/ModelGeneratorTest.php
@@ -215,9 +215,10 @@ class ModelGeneratorTest extends GiiTestCase
         }
     }
 
-    public function testGenerateClassName()
+    public function testGenerateClassNameForcingPsr()
     {
         $modelGenerator = new ModelGeneratorMock;
+        $modelGenerator->forcePsrClassNames = true;
 
         $tableNames = [
             'lower_underline_name' => 'LowerUnderlineName',
@@ -230,6 +231,32 @@ class ModelGeneratorTest extends GiiTestCase
             'lowerUcwordsName' => 'LowerUcwordsName',
             'lowername' => 'Lowername',
             'UPPERNAME' => 'Uppername',
+        ];
+
+        foreach ($tableNames as $tableName => $expectedClassName) {
+            $generatedClassName = $modelGenerator->publicGenerateClassName($tableName);
+            $this->assertEquals($generatedClassName, $expectedClassName);
+        }
+    }
+
+    public function testGenerateClassNameNoForcingPsr()
+    {
+        $modelGenerator = new ModelGeneratorMock;
+        $modelGenerator->forcePsrClassNames = false;
+
+        $tableNames = [
+            'lower_underline_name' => 'LowerUnderlineName',
+            'Ucwords_Underline_Name' => 'UcwordsUnderlineName',
+            'UPPER_UNDERLINE_NAME' => 'UPPERUNDERLINENAME',
+            'ABBRMyTable' => 'ABBRMyTable',
+            'lower-hyphen-name' => 'Lower-hyphen-name',
+            'Ucwords-Hyphen-Name' => 'Ucwords-Hyphen-Name',
+            'UPPER-HYPHEN-NAME' => 'UPPER-HYPHEN-NAME',
+            'CamelCaseName' => 'CamelCaseName',
+            'lowerUcwordsName' => 'LowerUcwordsName',
+            'lowername' => 'Lowername',
+            'UPPERNAME' => 'UPPERNAME',
+            'PARTIALUpperName' => 'PARTIALUpperName',
         ];
 
         foreach ($tableNames as $tableName => $expectedClassName) {
--- a/tests/generators/ModelGeneratorTest.php
+++ b/tests/generators/ModelGeneratorTest.php
@@ -215,10 +215,10 @@ class ModelGeneratorTest extends GiiTestCase
         }
     }
 
-    public function testGenerateClassNameForcingPsr()
+    public function testGenerateStandardizedCapitalsForClassNames()
     {
         $modelGenerator = new ModelGeneratorMock;
-        $modelGenerator->caseInsensitive = true;
+        $modelGenerator->standardizeCapitals = true;
 
         $tableNames = [
             'lower_underline_name' => 'LowerUnderlineName',
@@ -235,14 +235,14 @@ class ModelGeneratorTest extends GiiTestCase
 
         foreach ($tableNames as $tableName => $expectedClassName) {
             $generatedClassName = $modelGenerator->publicGenerateClassName($tableName);
-            $this->assertEquals($generatedClassName, $expectedClassName);
+            $this->assertEquals($expectedClassName, $generatedClassName);
         }
     }
 
-    public function testGenerateClassNameNoForcingPsr()
+    public function testGenerateNotStandardizedCapitalsForClassNames()
     {
         $modelGenerator = new ModelGeneratorMock;
-        $modelGenerator->caseInsensitive = false;
+        $modelGenerator->standardizeCapitals = false;
 
         $tableNames = [
             'lower_underline_name' => 'LowerUnderlineName',
@@ -261,7 +261,7 @@ class ModelGeneratorTest extends GiiTestCase
 
         foreach ($tableNames as $tableName => $expectedClassName) {
             $generatedClassName = $modelGenerator->publicGenerateClassName($tableName);
-            $this->assertEquals($generatedClassName, $expectedClassName);
+            $this->assertEquals($expectedClassName, $generatedClassName);
         }
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Yii Framework 2 gii extension Change Log
 -----------------------
 
 - Bug #327: Fixed bug in Model generator when $baseClass is an abstract class (rhertogh)
-- Enh #366: Optional PSR compliant class names for model generator (slinstj)
+- Enh #366: Option to allow standardized class names capitals in model generator (slinstj)
 
 
 2.0.7 May 3, 2018
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ Yii Framework 2 gii extension Change Log
 -----------------------
 
 - Bug #327: Fixed bug in Model generator when $baseClass is an abstract class (rhertogh)
-- Enh #366: Better class and file names for uppercase tables (slinstj)
+- Enh #366: Optional PSR compliant class names for model generator (slinstj)
 
 
 2.0.7 May 3, 2018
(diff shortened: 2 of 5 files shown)


rob006:
`ABBRMyTable` *is* PSR compliant. Name of property and description is misleading.

slinstj:
@rob006 you are right. StudlyCaps has no a fixed definition. On the other hand, `AbbrmyTable` can be PSR too, thanks to StudlyCaps again. 
Do you have any practical suggestion for this PR ? You are welcome ! Thnx

rob006:
I would rather create option like "case insensitive" which will lowercase table name (`MY_TABLE` -> `my_table`) and then generate class name like before (`MyTable`).

slinstj:
This not only is *already* done in this PR but this is *why* it was done ;-) 

But "Case insensitive" name is a really good. I will change it. Thnx.

If you both have more suggestions let me know plz.

slinstj:
One more point, when you said "like before (MyTable)" you need to know that it was not the case because a name like `MY_TABLE` were not converted to `MyTable` but to `MYTABLE`. Now we have this option exactly because this PR.

slinstj:
It is not so simple how it seems (I found it when developing). I will give you an exemple:

`strtolower('myTable')` results `mytable` .  Then `ucwords('mytable')`  results `Mytable`. Is this the desired result?

rob006:
If you have an option which is named "case insensitive", then IMO yes, it is desired behavior, since case should be ignored according to option name and description.

slinstj:
I would recommend you to read the issue 366, specially the openning message, please. I am pretty sure you have not understood what is the real problem here because, based in your argument, we would not need any additional option at all.

rob006:
> based in your argument, we would not need any additional option at all.

I think there is exactly the opposite. This additional option allows to make this really simple. Current overcomplicated solution is a result of your attempt to handle case from #366 and keep BC at the same time. This attempt failed. Since now we're introducing special setting for #366, we don't need to even try to keep BC when this option is enabled. Simple `strtolower()` will be enough to cover #366. And BC will be kept for everyone who does not explicitly check this option.

slinstj:
You are right (now that I understand what you mean)! I am sorry! I am at work but when I  get home I will check that but I think you are completely right. Thank you.

slinstj:
Ok guys, I tried using `Inflector::id2camel(strtolower($schemaName.$className), '_');` but with no success. As 'with no success' I mean the result is not the one initially expected. 
My inital intention when using this new option was to achieve this:
```php
$tableNames = [
            // table name => expected name,
            'lower_underline_name' => 'LowerUnderlineName',
            'Ucwords_Underline_Name' => 'UcwordsUnderlineName',
            'UPPER_UNDERLINE_NAME' => 'UPPERUNDERLINENAME',
            'ABBRMyTable' => 'ABBRMyTable',
            'lower-hyphen-name' => 'Lower-hyphen-name', // ** test fails
            'Ucwords-Hyphen-Name' => 'Ucwords-Hyphen-Name', // ** test fails
            'UPPER-HYPHEN-NAME' => 'UPPER-HYPHEN-NAME', // ** test fails
            'CamelCaseName' => 'CamelCaseName', // ** test fails
            'lowerUcwordsName' => 'LowerUcwordsName', // ** test fails
            'lowername' => 'Lowername',
            'UPPERNAME' => 'UPPERNAME',
            'PARTIALUpperName' => 'PARTIALUpperName',
        ];
```
Any hyphenized or camelized names does not work, only the first char is capitalized.

**However**, I managed to simplify the things while yet achieving the desired results. I changed this:
```php
str_replace(' ', '', ucwords(implode(' ', explode('_', strtolower(str_replace(' ', '_', Inflector::camel2words($schemaName.$className)))))));
```
by this (infinitely more simple!):
```php
Inflector::camelize(Inflector::camel2words($schemaName.$className));
```
All tests passes again.

Please, let me know what do you think.

rob006:
If you want to support hyphens as a word separator in table name, all you need is `strtr($schemaName . $className, ['-' => '_'])` before lowercasing table name.

> 
    'CamelCaseName' => 'CamelCaseName', // ** test fails
    'lowerUcwordsName' => 'LowerUcwordsName', // ** test fails

These tests **should** fail. If you want to keep case, you should not use `caseInsensitive` option.

slinstj:
> If you want to support hyphens as a word separator in table name, all you need is strtr($schemaName . $className, ['-' => '_']) before lowercasing table name.

It could be, no problem. I just dont see why this:
```Inflector::id2camel(strtolower(strtr($schemaName . $className, ['-' => '_'])), '_');```
... is better than this:
```Inflector::camelize(Inflector::camel2words($schemaName.$className))```
... since we had already a dependency on Inflector anyway.

> These tests should fail. If you want to keep case, you should not use caseInsensitive option.

I see that we are thinking of differente ways. I understand `caseInsenstive` like "Ok, it is not important to me if my classes capitalizing are identical to the tables capitalizing or not" or, in other words, "I dont worry how Yii will make my classes names capitalizing". So that Yii will be free to StudlyCaps (or whatever) the names.

But just to avoid more misunderstoods, I see two ways:

1. Change the option name to "StudlyCaps" name or something like (although StudlyCaps are not standardized, it is the most of time a `ucfirst('camelCaseWord')`);
2. Or, if you dont think it would be useful we have a new option like this (and maybe it really will not be for everyone) I can just convert it to an extension. That would be enough to me (and maybe to others). However, i just dont think that generate a class name like `Mymaintable` or `MYMAINTABLE` is better than `MyMainTable`.

Now, I let it up to you decide. :wink: :+1:

**Comment by slinstj** [comment]
Unfortunatelly, althougn "simpler to maintain and understand", it does not work like intended:
```shell
1) yiiunit\gii\generators\ModelGeneratorTest::testGenerateClassNameForcingPsr
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'Camelcasename'
+'CamelCaseName'
```
and unfortunatelly it breaks the current implementation too:
```shell
2) yiiunit\gii\generators\ModelGeneratorTest::testGenerateClassNameNoForcingPsr
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'LowerHyphenName'
+'Lower-hyphen-name'
```

rob006:
Tests are green now ;).

slinstj:
Guy, you changed the tests! Of course it works now. Lol... Ok, but could you explain why we cant have names like SomeNameTable instead somenametable?

rob006:
> Ok, but could you explain why we cant have names like SomeNameTable instead Somenametable?

You can - just don't use `caseInsensitive` option.

tests/generators/ModelGeneratorTest.php
```diff
@@ -215,9 +215,10 @@ public function testRules($tableName, $fileName, $rules)
         }
     }
 
-    public function testGenerateClassName()
+    public function testGenerateClassNameForcingPsr()
```
samdark: It's not PSR. Test name should be changed.

rob006:
There are tests that prove that you can: https://github.com/yiisoft/yii2-gii/pull/370/files#diff-838c9ed27dc7fb7a7aa0b5af2ccafe90R256

Or you need to be more specific what table name do you have and what model name do you expect.

slinstj:
@samdark, I will make the change and you can go ahead and merge if you want. But, since we are user/contributors to this opensource project, I think it is so fair we have some justify when something is approved or changed or refused.

I really don't agree that we should have a table named "CamelCaseName" generating a class "Camelcasename". Why, in the world, would somebody name your table like "ProductDetails" just to have a class "Productdetails"? In the own Yii2 repository we can't find this (and I think we can't in any other)! 

I really respect all the work is done here in Yii2 and I know that I have so much to learn from you (including from @rob006). You are really great programmers. Someday I hope to be in your levels. But my question about **why** we would not have this approach had not been answered and now you both want it to be merged without answer. It is just not much fair. To say that the code is "complex" is not a justify. You **must** have a valid justify to preffer names like "Productdetails".

slinstj:
@rob006, *using the new approach you suggested*: 
```php
        $name = strtr($schemaName . $className, [
            '-' => '_',
        ]);

        if ($this->caseInsensitive) {
            $name = strtolower($name);
        }

        return $this->classNames[$fullTableName] = Inflector::id2camel($name, '_');
```
Results when `caseInsensitive` is enabled:
```shell
# table name - expected name - generated name
CamelCaseName - CamelCaseName - Camelcasename # fails
lowerUcwordsName - LowerUcwordsName - Lowerucwordsname # and fails
```

rob006:
@slinstj I repeat again: If you want to keep the case of your table name, then **do not use `caseInsensitive` option**. This new switch is **optional**. Why do you want to use it if you want to get `CamelCaseName` from `CamelCaseName`? Why do you expect that these two scenarios should work in the same way? It does not make any sense...

samdark:
OK, seems this isn't going anywhere. Let's start over.

1. We should keep it backwards compatible by default.
2. New mode should be OK for the following ideal test set:

```php
'lower_underline_name' => 'LowerUnderlineName',
'Ucwords_Underline_Name' => 'UcwordsUnderlineName',
'UPPER_UNDERLINE_NAME' => 'UpperUnderlineName',
'ABBRMyTable' => 'ABBRMyTable',
'lower-hyphen-name' => 'LowerHyphenName',
'Ucwords-Hyphen-Name' => 'UcwordsHyphenName',
'UPPER-HYPHEN-NAME' => 'UpperHyphenName',
'CamelCaseName' => 'CamelCaseName',
'lowerUcwordsName' => 'LowerUcwordsName',
'lowername' => 'Lowername',
'UPPERNAME' => 'Uppername',
'PARTIALUpperName' => 'PARTIALUpperName',
```

From the test set I'm getting the following rules:

1. `-` and `_` are clearly word separators. So split into array of words first then apply following for each word.
2. All-uppercased words should be lowercased first.
3. First letter should be always uppercased.
4. Mixed case words should be left as is.

samdark:
@rob006, @slinstj what do you think?

rob006:
I still don't see any sense in forcing the same behavior i two separated scenarios. This makes code more complicated and harder to maintain, behavior becomes harder to explain and understand, and finally, this solution is less flexible. What if I **want to** get `Camelcasecame` class name from `CamelCaseName` table name? With "my" solution I have a choice - I can use `caseInsensitive` option and get what I want (`Camelcasecame`) or don't use `caseInsensitive` option and get what you want (`CamelCaseName`). I do not understand these efforts and complications just to take away this choice from me and not support my usecase...

slinstj:
> If you want to keep the case of your table name, then do not use caseInsensitive option

This option is intended to  when generating *multiple* tables (using asterisc) and this is why this option disappears when not using asterisc in table name. If I have tables mixing `UPPERCASE` and `StudlyCaps` names and I am using asterisc to generate classes and I didn't enabled `caseInsensitive` everything will be kept but if I want to *standardize* everything to StudlyCaps when generating those *multiple* tables it should be enabled. (I think we could be having some type of language barrier)

BTW, `standardizeNames` seems interesting too.

Samdark, it is fine to me since as far as I can see it is already done.

> - and _ are clearly word separators. 

Already done when using new option (for BC).

> split into array of words first then apply following for each word.

Already being done too.

> All-uppercased words should be lowercased first.

Done too and is the reason that I use the "complex" code (`ctype_uppercase()`) to check *all* before;

> First letter should be always uppercased.

Done too.

> Mixed case words should be left as is.

This is the default behavior,  if *not* using the new option.

Besides that, I am stopping here... I dont think this simple option would deserve so much discussion. I am fine anyway (no hurts at all) and hope to contribute in others opportunities. :+1:

samdark:
To cover custom cases we may introduce a callback-option there.

samdark:
@slinstj you mean done before @rob006 commit?

slinstj:
Yeah, before his changings.

samdark:
@rob006 I think we should adopt @slinstj solution. How about that and adding a callback that allows customization?

rob006:
@samdark Honestly, I don't care anymore. But I'm pretty sure that you will not be able to describe this behavior in any simple way, and it will return garbage for some cases (like `IAmATable`) since you're trying to detect and ignore the case of tables at the same time. Basically, you're trying to create a dedicated solution to support individual use case and ignore everything else.

samdark:
IAmATable should stay as is and it won't be garbage, it will do exactly that.

rob006:
@samdark  According to https://github.com/yiisoft/yii2-gii/issues/366#issuecomment-395903464 and description from this PR, is should be changed to `IamAtable`. If it is not true, then why we have this option in the first place?

samdark:
Nah, that one is about breaking old behavior.

slinstj:
@samdark, I roll back the code to before the changings and I updated it a bit...

So, this PR is about to *allow* generate class names in a *standardized* form **only** when generating *multiple* tables (using *). This is why I changed the option name to `Standardize Names`. Let me know what do you think. `caseInsensitive` would probably have double understanding just like we saw here during discussions.

Please, double check the code and let me know if something should be changed.

samdark:
Merged. Thank you!

PR #380, opened 2018-09-14
Force call find method in static instead of itself

desenvolvedorindie:
Makes it possible refer itself in inherited class.
In the grid, you can receive the search class generated in the $model parameter
Call itself find method, if exist
It possible create fields in itself and auto fill in find()

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no


rob006:
This may be undesirable, since data provider will return search model instances instead of model.

cebe:
This may result in wrong search results, not a good idea to do in general.

PR #381, opened 2018-09-16
remove hard coded Id field

shujaattariq:
hard coded $id field could be $_id for mongodb, changed code to use the actual PK

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |


cebe:
the name of the action parameter does not affect the name in the database, can you explain more about why you think this change is needed?

Issue #260, opened 2017-03-15
test bootstrap 4 compatibility and adjust version constraint if possible


cebe:
https://github.com/yiisoft/yii2-gii/pull/259#issuecomment-286379114

PR #390, opened 2018-11-30
Create (bootstrap)-independent version

simialbi:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌ 
| New feature?  | ✔️ 
| Breaks BC?    | ❌ 
| Tests pass?   | ✔️ 
| Fixed issues  | #260, PR #384 

Like https://github.com/yiisoft/yii2-debug/pull/340 for yii2-debug the same for yii2-gii.

Work in progress demo: http://demo.karlen.li/web/gii
For model generation: Table `yii_user` exists


samdark:
Great to see you working on gii as well!

Some feedback (I know it's work in progress but still). Based on demo:

1. It looks better w/o paddings of the main page.
2. Header looks different. Not as good as it was.
3. Text sizes and styles are different.
4. Pressing "Preview" gives you nothing.

simialbi:
@samdark should I rewrite [gii.js](https://github.com/simialbi/yii2-gii/blob/master/src/assets/js/gii.js) to vanilla?

cebe:
> @samdark shall I rewrite [gii.js](https://github.com/simialbi/yii2-gii/blob/master/src/assets/js/gii.js) to vanilla?

I don't think this is needed, the frontend of Gii can use jquery without conflicting with other code.
When writing extensions for Gii (custom generators) we can use jquery to optimize the frontend form, having to use native JS does not bring any benefit in this case.

simialbi:
I think it's ready for a review now. I removed typeahead.js beacause of the discussion of [jquery dependance](https://github.com/yiisoft/yii2-gii/pull/390#issuecomment-443455503). I replaced it with native [datalists](https://www.w3schools.com/tags/tag_datalist.asp). It's supported in the most modern major browsers: https://caniuse.com/#search=datalist. I looked for an alternative and e.g. [Awesomeplete](https://github.com/leaverou/awesomplete/) is a lightweight one.

As soon as you reviewed I'm going to clean up the code **if you agree** (globally apply yii2 code style, add phpdoc etc.)

samdark:
There seems to be an issue with preview buttons:

buttons

Looks like it could be solved with `background-repeat: no-repeat`.

schmunk42:
Sorry, I may have asked this before, but why are all scss bootstrap files included in the source?
I thought this is bootstrap-free :)

simialbi:
It's bootstrap **independent**, not bootstrap **less**. Bootstrap is bundled to solve problems like this:
https://github.com/yiisoft/yii2-bootstrap4/issues/50

schmunk42:
The issue linked above won't be a problem with `npm` btw. or at least it would be different, since you can install two versions when using native tooling - but not with composer. But this is more Yii 3.x related. CC: @samdark @cebe 

I don't like the idea of bundling a whole CSS framework with a module, although I see that there has to be some style. And for Yii 2.x this might be the way to go.

But we should think about this for https://github.com/yiisoft/yii-gii/

simialbi:
I know it's not the best solution, but the only alternative I see is to build a dependency free (as far as possible) lightweight yii css framework for cases like this or [yiisoft/yii2-debug](https://github.com/yiisoft/yii2-debug).

samdark:
It's fine for Yii 2.x and is close to what we need in Yii 3.0.

schmunk42:
@samdark You want to bundle Bootstrap 4 for Yii 3.x for gii, debug and for yii-bootstrap4? So it would be included 3 times probably in different versions?

simialbi:
@schmunk42 you still have to consider the following: gii and debug are development modules you don't need in production. So mabye it's not as tragic as it seems

samdark:
@schmunk42 we may do it. As @simialbi said, there's nothing wrong with it. Still, ideally we'd do our own simplistic CSS.

schmunk42:
> you have to consider the following: gii and debug are development modules you don't need in production. So mabye it's not as tragic as it seems

While I agree for debug, I had use-cases with gii, where I wanted to generate code on a website, ie. think about an online extension generator.

> there's nothing wrong with it. Still, ideally we'd do our own simplistic CSS.

I totally agree for Yii 2.x, but we should consider the use-case in general, that you want to install two different version of a CSS framework; if you have something like AdminLTE you might also be tied to certain JS libs.

For 3.x we should try to use minimal custom CSS and don't try to find workarounds (in core modules) from the beginning.

samdark:
Merged. Thank you very much!

PR #394, opened 2018-12-16
Return to previous page after deletion

gladunrv:
For example, after deleting from the page
/index?page=2
will redirect to this page again

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | my issues

### DIFF
--- a/src/generators/crud/default/controller.php
+++ b/src/generators/crud/default/controller.php
@@ -147,7 +147,12 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
     {
         $this->findModel(<?= $actionParams ?>)->delete();
 
-        return $this->redirect(['index']);
+        list($actionID) = Yii::$app->createController(parse_url(Yii::$app->request->referrer, PHP_URL_PATH));
+        if ($actionID === 'index') {
+            return $this->redirect(Yii::$app->request->referrer);
+        } else {
+            return $this->redirect(['index']);
+        }
     }
 
     /**


machour:
@gladunrv I agree with the idea but I'm not really found of the solution provided here.

We may want to use `Url::remember()` & `$this->goBack()` instead, and handle more cases, like when you edit from the index page, and are redirected to the view page instead of getting back to index.

I would:

- Generate calls to `Url::remember()` in `actionView/Index`
- In `actionUpdate/Edit`, use $this->goBack()
- In `actionDelete`: If previous url is "view", then redirect to "index", else, `goBack()`

What do you think? 

NB: This will use sessions instead of the HTTP referer..

schmunk42:
@machour We tried similar things in https://github.com/schmunk42/yii2-giiant, especially with `Url::remember()`, but this created a lot of problems when working with multiple tabs.
Moreover if you have a delete button in a frontend page (as admin), you get redirected to a non-existent page, when you go back.
Long story short, we removed all that magic and sticked with minimal logic. The proposed solution looks good to me.

src/generators/crud/default/controller.php
```diff
@@ -147,7 +147,12 @@ public function actionDelete(<?= $actionParams ?>)
     {
         $this->findModel(<?= $actionParams ?>)->delete();
 
-        return $this->redirect(['index']);
+        list($actionID) = Yii::$app->createController(parse_url(Yii::$app->request->referrer, PHP_URL_PATH));
```
rob006: Why you're assuming that `parse_url(Yii::$app->request->referrer, PHP_URL_PATH)` will return correct route?

rob006:
This is looks brittle at best. Referrer is not always available, and assumption that URL path is a correct route  and `index` action is somehow special, looks to fragile to me. If we really need something like this, we should check for `$_GET['return']` and pass return URL directly in delete URL.

samdark:
Passing previous URL in a GET variable sounds like slightly more reliable solution. @gladunrv would you like to adjust the pull request?

schmunk42:
How about using `$_GET['return_url']`?

PR #395, opened 2019-01-02
Made yii\gii\CodeFile indepdendent of controller context

cebe:
do not apply `$newDirMode` and `$newFileMode` if module is not available.

Code uses default umask in that case.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Yii Framework 2 gii extension Change Log
 -----------------------
 
 - Enh #390, Bug #260: Create (bootstrap)-independent version (simialbi)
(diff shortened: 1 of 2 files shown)


samdark:
What's the use case?

cebe:
I spotted this when running unit tests for the gii Generator in https://github.com/cebe/yii2-openapi tests fail with setting file permission of 0 if gii module is not mocked. 

Imo the class should not rely on gii module being there, this makes testing much easier as you do not need to mock a whole yii app ffor testing a gii generator.

samdark:
Merged. Thank you!

PR #399, opened 2019-02-14
Closes #292. Add singularize class names option.

### LINKED ISSUE #292: Small suggestion: Option to singularize class names
I prefer to have all class names as singular since each object is one item.  I think this is quite common as well.

Unfortunately, db tables sometimes tend to be plural, since they are a collection of several.

Renaming these manually after using Gii is a bit of a pain because of all the relations, query classes, and such.

alexkart:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #292

### DIFF
--- a/tests/generators/ModelGeneratorTest.php
+++ b/tests/generators/ModelGeneratorTest.php
@@ -264,4 +264,45 @@ class ModelGeneratorTest extends GiiTestCase
             $this->assertEquals($expectedClassName, $generatedClassName);
         }
     }
+
+    public function testGenerateSingularizedClassNames()
+    {
+        $modelGenerator = new ModelGeneratorMock;
+        $modelGenerator->singularize = true;
+
+        $tableNames = [
+            'clients' => 'Client',
+            'client_programs' => 'ClientProgram',
+            'noneexistingwords' => 'Noneexistingword',
+            'noneexistingword' => 'Noneexistingword',
+            'children' => 'Child',
+            'good_children' => 'GoodChild',
+            'user' => 'User',
+        ];
+
+        foreach ($tableNames as $tableName => $expectedClassName) {
+            $generatedClassName = $modelGenerator->publicGenerateClassName($tableName);
+            $this->assertEquals($expectedClassName, $generatedClassName);
+        }
+    }
+
+    public function testGenerateNotSingularizedClassNames()
+    {
+        $modelGenerator = new ModelGeneratorMock;
+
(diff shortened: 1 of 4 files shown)


tests/generators/ModelGeneratorTest.php
```diff
@@ -264,4 +264,45 @@ public function testGenerateNotStandardizedCapitalsForClassNames()
...
     }
+
+    public function testGenerateSingularizedClassNames()
+    {
+        $modelGenerator = new ModelGeneratorMock;
+        $modelGenerator->singularize = true;
+
+        $tableNames = [
+            'clients' => 'Client',
+            'client_programs' => 'ClientProgram',
```
rob006: What about `clients_programs`?

schmunk42: And `some_data` ... AFAIR becomes `some_datum`?

alexkart: > What about `clients_programs`?

What about it? As you can see `client_programs` becomes `ClientProgram`. Or you mean something else?

alexkart: > And `some_data` ... AFAIR becomes `some_datum`?

In this case no, it uses `\yii\helpers\Inflector` and it supports the following exceptions https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseInflector.php#L109

schmunk42: The default is `false` so this is fine. But there might be some cases you won't expect; I just wanted to point that out.

alexkart: > The default is `false` so this is fine. But there might be some cases you won't expect; I just wanted to point that out.

Yes, it may not cover all possible cases, it uses * `\yii\helpers\Inflector` and it supports the following rules and exceptions https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseInflector.php#L63

alexkart:
Is eveyhing OK with the PR?

schmunk42:
I think we need to switch to phpunit from vendor to make 7.2 pass

machour:
Already done in https://github.com/yiisoft/yii2-gii/commit/505c92d978ad06c763d0db536e99580a732d8e94, I guess you need to merge changes from master

samdark:
Merged. Thank you!

PR #403, opened 2019-03-17
use `::class` instead className()

spanjeta:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  | comma-separated list of tickets # fixed by the PR, if any
Minor improvement.  use `::class` instead className() in model generator. Its important because we generate lot of code and className() is depericated. So this change must be done early.


machour:
Sorry but we're maintaining PHP 5.4 compatibility, and `::class` was introduced in 5.5, so we cannot merge this.

PR #417, opened 2019-11-12
bugfix: fixed issue where RTL implementation for foreign keys (e.g. c…

NickvdMeij:
bugfix: fixed issue where RTL implementation for foreign keys (e.g. column name `id_user` for a foreign key) causes problems with LTR tables names (e.g. foreign key to `IdentityProvider` table becomes relation function `getEntityProviders`)

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,9 @@ Yii Framework 2 gii extension Change Log
 2.1.2 October 08, 2019
 ----------------------
 
-- Bug #417: Fixed issue where RTL implementation for foreign keys causes problems with LTR tables names
+- Bug #417: Fixed issue where RTL implementation for foreign keys causes problems with LTR tables names (NickvdMeij)
 - Bug #413: Controller Generator produces invalid alias when namespace starts with backslash (cebe)
+
 
 2.1.1 August 13, 2019
 ---------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,8 @@ Yii Framework 2 gii extension Change Log
 2.1.2 October 08, 2019
 ----------------------
 
+- Bug #417: Fixed issue where RTL implementation for foreign keys causes problems with LTR tables names
 - Bug #413: Controller Generator produces invalid alias when namespace starts with backslash (cebe)
-
 
 2.1.1 August 13, 2019
 ---------------------
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -719,8 +719,8 @@ class Generator extends \yii\gii\Generator
         if (!empty($key) && strcasecmp($key, 'id')) {
             if (substr_compare($key, 'id', -2, 2, true) === 0) {
                 $key = rtrim(substr($key, 0, -2), '_');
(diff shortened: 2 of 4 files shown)


samdark:
That isn't a trivial thing. Is it possible to cover it with test so it doesn't break anymore?

NickvdMeij:
Hey @samdark 

Like you asked, I implemented a unit test to set these relations with RTL syntax (id_user instead of user_id) and also the situation where it gave problems (in identity_provider_id foreign key in an organization table). However, I could not get the tests to work locally due to the following error:

`The value of autoloader is specified, but file doesn't exist ''`

even though I did run the composer install and the autoloader file has been generated. Sadly due to time constraints, I was unable to invest more time to sort this out..

Furthermore, I looked at the test results from the previous build, and saw there was one empty (risky) test which I removed (empty function has no use in source code?)

Do you mind looking at the current test implementation (and assuming you got everything running locally) and test is for me? 

Thanks in advance,

Nick van der Meij
MyCademy B.V.

NickvdMeij:
@samdark I fixed the syntax errors in the SQL, seems like the tests run as usual in travis CI now. From what i can see, there are still some (minor) problems with the tests, however the relation "getIdentityProviders()" is now generated correctly (was previously "getEntityProviders()" as mentioned in the first description of the PR).

samdark:
There are still some breaks: https://travis-ci.org/yiisoft/yii2-gii/jobs/611346074?utm_medium=notification&utm_source=github_status

NickvdMeij:
My bad, looking at result of the tests before and after my implementation, I realized the inverseOf's weren't generated at all (also not for the test that was already implemented). Removed the inverseOf's in my tests as well

samdark:
OK. Less failures now but still it seems something is broken: https://travis-ci.org/yiisoft/yii2-gii/jobs/611775091

NickvdMeij:
@samdark I just commited the (from what I see) last fixes for the tests, my assumption (since I can still not run the tests locally) is that everything should be green now. However, latest commit hasn't been ran by TravisCI. Is there a way to manually trigger this?

NickvdMeij:
@samdark last but not least, all tests pass.
Sorry it took a while. Due to the fact that I was unable to test locally there were more commits then there should be. Hopefully this won't be a big problem ;-).

samdark:
Not a problem at all.

samdark:
Would you please add a line to changelog?

NickvdMeij:
@samdark  Added changelog line ;-) Any indication when this will be merged into the master branch? Would love to use this package again instead of my own fork in my php project ;)

samdark:
@NickvdMeij soon. Maybe even today.

samdark:
Merged. Thank you!

NickvdMeij:
@samdark no you thank you with this awesome generator =D you don't wanna know how much time this has saved us in the past few years. Glad I had the change to help you guys out for a change :) Keep doing the good work, really appreciated!

Issue #402, opened 2019-02-26
The generator makes all integer columns as […, 'default', 'value' => null] even column set as not null


Mirocow:
### What steps will reproduce the problem?

We have some problem with pgsql generating model rules
The generator makes all integer columns as […, 'default', 'value' => null] even column set as not null

### What do you get instead?

Replace code from (https://github.com/yiisoft/yii2-gii/blob/master/src/generators/model/Generator.php#L358)

```php
if ($driverName === 'pgsql' && $type === 'integer') {
    $rules[] = "[['" . implode("', '", $columns) . "'], 'default', 'value' => null]";
}
```

into

```php
if ($driverName === 'pgsql' && $type === 'integer') {
    $defaultColumns = isset($types['required']) ? array_diff($columns, $types['required']) : $columns;
    if ($defaultColumns) {
        $rules[] = "[['" . implode("', '", $defaultColumns) . "'], 'default', 'value' => null]";
    }
}
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.16?
| PHP version      |7.2 
| Operating system | Debian

alexkart:
I've verified this and confirm that it generates the same default rule null and not null columns.
But, does anyone know what's the purpose of specifying a default value for integers in postgresql?
I mean why is this block of code needed?
```php
if ($driverName === 'pgsql' && $type === 'integer') {
    $rules[] = "[['" . implode("', '", $columns) . "'], 'default', 'value' => null]";
}
```

Issue #335, opened 2018-02-19
Gii sets wrong model rule for integer field with default value


kjusupov:
### What steps will reproduce the problem?

```SQL
create table parent (id serial primary key, code character varying(5));
insert into parent (code) values ('one');
insert into parent (code) values ('two');
```

Next, create `child` table with FK pointing to `parent.id`

```SQL
create table child (
    id serial primary key, 
    parent_id integer references parent(id) not null default 1, 
    code character varying(5)
);

insert into child (parent_id, code) values (1, 'c 1');
insert into child (parent_id, code) values (2, 'c 2');
insert into child (code) values ('c 3');   -- parent_id will be defaulted to `1`
```

Next, generate models via `Gii/model`

### What is the expected result?

```php
    /**
     * @inheritdoc
     */
    public function rules()
    {
        return [
            [['parent_id'], 'default', 'value' => 1], <--- this is what I expected
            [['parent_id'], 'integer'],
            [['code'], 'string', 'max' => 5],
            [['parent_id'], 'exist', 'skipOnError' => true, 'targetClass' => Parent::className(), 'targetAttribute' => ['parent_id' => 'id']],
        ];
    }
```
### What do you get instead?

```php
    /**
     * @inheritdoc
     */
    public function rules()
    {
        return [
            [['parent_id'], 'default', 'value' => null], <--- this is what I get
            [['parent_id'], 'integer'],
            [['code'], 'string', 'max' => 5],
            [['parent_id'], 'exist', 'skipOnError' => true, 'targetClass' => Parent::className(), 'targetAttribute' => ['parent_id' => 'id']],
        ];
    }
```

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.13.1
| Gii | 2.0.6.0
| PHP version      | 7.0
| PostgreSQL version      | PostgreSQL 9.5.9 on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609, 64-bit
| Operating system | Ubuntu 16.04 LTS

slinstj:
This was added for sake of #224.

kjusupov:
I submitted a PR, not sure how to test other than what I did below (manual tests):

My test DB tables:
```sql
create table brand (id serial primary key, code character varying(20));
create table country (id serial primary key, code character varying(20));
create table region (id serial primary key, code character varying(20));
[...]

PR #420, opened 2019-11-16
To model rules add default values

uldisn:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

To model rules add default values.
Example:

```php
    /**
     * {@inheritdoc}
     */
    public function rules()
    {
        return [
            [['name', 'address', 'profile_id'], 'default', 'value' => null],
	    [['status'], 'integer', 'default', 'value' => 0],
            [['email'], 'required'],
            [['address'], 'string'],
            [['status', 'profile_id'], 'integer'],
            [['email', 'name'], 'string', 'max' => 128],
        ];
    
```


schmunk42:
I think this should either be enabled by a parameter or done in a 2.2.x release.

uldisn:
Realy is bug! See description.

src/generators/model/Generator.php
```diff
@@ -335,6 +335,49 @@ public function generateRules($table)
     {
         $types = [];
         $lengths = [];
+        $rules = [];
+        $driverName = $this->getDbDriverName();
+
+        /**
+         * Default values
```
samdark: No need for this comment. It's clear that these are default values because variable names are self-explanatory.

src/generators/model/Generator.php
```diff
@@ -335,6 +335,49 @@ public function generateRules($table)
...
+         */
+        $columnsDefaultNull = [];
+        $columnsDefaultValues = [];
+        foreach ($table->columns as $column) {
+            if (in_array($driverName, ['mysql', 'sqlite'], true)) {
```
samdark: `in_array` could be moved out of `foreach`.

src/generators/model/Generator.php
```diff
@@ -335,6 +335,49 @@ public function generateRules($table)
...
+                        case Schema::TYPE_TINYINT:
+                        case Schema::TYPE_BOOLEAN:
+                        case Schema::TYPE_FLOAT:
+                        case Schema::TYPE_DOUBLE:
+                        case Schema::TYPE_DECIMAL:
+                        case Schema::TYPE_MONEY:
+                            $defaultValue = $column->defaultValue;
+                            break;
+                        default:
+                            $defaultValue = '\'' . $column->defaultValue . '\'';
```
samdark: What if the value itself contains `'`?

src/generators/model/Generator.php
```diff
@@ -335,6 +335,49 @@ public function generateRules($table)
...
+
+        /**
+         * Default values
+         */
+        $columnsDefaultNull = [];
+        $columnsDefaultValues = [];
+        foreach ($table->columns as $column) {
+            if (in_array($driverName, ['mysql', 'sqlite'], true)) {
+                /**
+                 * text default values quote
```
samdark: It is better to remove this comment.

src/generators/model/Generator.php
```diff
@@ -335,6 +335,54 @@ public function generateRules($table)
...
+                        case Schema::TYPE_TIMESTAMP:
+                            if(strtoupper($column->defaultValue) === 'CURRENT_TIMESTAMP'){
+                                $defaultValue = 'date(\'Y-m-d H:i:s\')';
+                            }else{
+                                $defaultValue = '\'' . addslashes($column->defaultValue) . '\'';
+                            }
+                            break;
+
+                        default:
+                            $defaultValue = '\'' . addslashes($column->defaultValue) . '\'';
```
samdark: It won't work https://stackoverflow.com/questions/1542627/escaping-column-names-in-pdo-statements

samdark: It's better to use `$db->getSchema()->quoteValue($column->defaultValue)`.

schmunk42:
I don't get it from the description, but apparently @samdark does, so it's fine :smile:

samdark:
Thinking more about it... @uldisn what's the use case for having explicit `default` rules if they match database ones? As far as I know, database silently uses default when value is not provided anyway...

uldisn:
Sometimes rules is fired, if not set default values. For example, if field required ( not null), in database is defined default value and in model not set. Then  rule required fired. 
For optimal indexing not null is very actual, but usually using default  CURRENT_TIMESTAMP fired on require rule.

samdark:
Isn't it better not to generate `required` rule in this case?

uldisn:
But my view is mirroring full table requirements in model. That allow in validation rules trust on default values. Currently my know problem is with require,  but other problems can occur.

For tuning masters yes. They can do it.

Additionally i see follow improvements:
 - default value define as model constant
 - create method "loadDefaultValues()" - useful for new record before  showing in form.

samdark:
> Currently my know problem is with require, but other problems can occur.

Any examples?

> create method "loadDefaultValues()" - useful for new record before showing in form.

https://www.yiiframework.com/doc/api/2.0/yii-db-activerecord#loadDefaultValues()-detail is already there and works without any extra rules.

uldisn:
On issue #335 is problem

samdark:
#335 seems to have another pull request https://github.com/yiisoft/yii2-gii/pull/421/files

uldisn:
Solutions:
a) database fields with not null and default value not include in require rule
* fired only on like #335 issue described situation
* minimal and optimal code
b) create full list rules for default values
* work correctly in any situation
* huge code
c) create full list rules for default values excluding null default value 
* work correctly in any situation
* no so huge code

tests/bootstrap.php
```diff
@@ -8,8 +8,8 @@
 $_SERVER['SCRIPT_NAME'] = '/' . __DIR__;
 $_SERVER['SCRIPT_FILENAME'] = __FILE__;
 
-require_once(__DIR__ . '/../vendor/autoload.php');
-require_once(__DIR__ . '/../vendor/yiisoft/yii2/Yii.php');
+require_once('../../autoload.php');
+require_once('../yii2/Yii.php');
```
samdark: Why changing it?

uldisn:
Generation default value rules cover follow problems:

- Gii sets wrong model rule for integer field with default value  #335 #421
- The generator makes all integer columns as […, 'default', 'value' => null] even column set as not nul #402

Only need adapt for PostgrSQL

WinterSilence:
that's magic in yii confuse me - why not define properties with defaults from db table?
~~~php
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `contract_number` varchar(50)  NOT NULL DEFAULT 'NoNumber' COMMENT 'Number',
  `contract_date` date DEFAULT NULL COMMENT 'Contract date',
  `contract_place` varchar(250) DEFAULT NULL COMMENT 'Contract place',
  `notes` text COMMENT 'Notes',
class Contact extends ActiveRecord
{
    /**
     * @var int
     */
    public $id;

    /**
     * @var string Number
     */
    public $contract_number = 'NoNumber';
    ...
}
~~~

src/generators/model/Generator.php
```diff
@@ -330,11 +330,64 @@ public function generateLabels($table)
...
+                        case Schema::TYPE_DOUBLE:
+                        case Schema::TYPE_DECIMAL:
+                        case Schema::TYPE_MONEY:
+                            $defaultValue = $column->defaultValue;
+                            break;
+
+                        case Schema::TYPE_DATETIME:
+                        case Schema::TYPE_TIMESTAMP:
+                            if(strtoupper($column->defaultValue) === 'CURRENT_TIMESTAMP'){
+                                $defaultValue = 'date(\'Y-m-d H:i:s\')';
```
WinterSilence, line 367: mssql/sql server: 
> DEFAULT definitions cannot be created on columns with a **timestamp** data type or columns with an IDENTITY property.

also it must be `$defaultValue = 'time()';`

uldisn:
> that's magic in yii confuse me - why not define properties with defaults from db table?
> 
> ```
> class Contact extends ActiveRecord
> {
>     /**
>      * @var int
>      */
>     public $id;
> 
>     /**
>      * @var string Number
>      */
>     public $contract_number = 'NoNumber';
> 
>     ...
> }
> ```
> 
> ```
> /**
>  * @var {type} {comment}
>  */
> public ${column} ={default};
> ```



> that's magic in yii confuse me - why not define properties with defaults from db table?
> 
> ```
> class Contact extends ActiveRecord
> {
>     /**
>      * @var int
>      */
>     public $id;
> 
>     /**
>      * @var string Number
>      */
>     public $contract_number = 'NoNumber';
> 
>     ...
> }
> ```
> 
> ```
> /**
>  * @var {type} {comment}
>  */
> public ${column} ={default};
> ```

For example, if default value is timestamp, it cannot set to property.

src/generators/model/Generator.php
uldisn: For timestamp column, if default value not set, fired on require rule.

WinterSilence: @uldisn 
~~~php
if (strtoupper($column->defaultValue) === 'CURRENT_TIMESTAMP') {
     if ($driverName === 'mssql') {
          $defaultValue = $this->generateString($column->defaultValue);
     } else {
        $defaultValue = 'time()';
    }
} 
~~~

WinterSilence:
@uldisn why? timestamp is integer

src/generators/model/Generator.php
uldisn: On MSSQL fired rule, who validate actual field value - time format.

src/generators/model/Generator.php
```diff
@@ -330,11 +330,64 @@ public function generateLabels($table)
...
+     * @throws NotSupportedException
      */
     public function generateRules($table)
     {
         $types = [];
         $lengths = [];
+        $rules = [];
+        $driverName = $this->getDbDriverName();
+
+        if (in_array($driverName, ['mysql', 'sqlite'], true)) {
```
WinterSilence, line 342: why only that's drivers?

src/generators/model/Generator.php
```diff
@@ -330,11 +330,64 @@ public function generateLabels($table)
...
+
+            $db = $this->getDbConnection();
+            $columnsDefaultNull = [];
+            $columnsDefaultValues = [];
+
+            foreach ($table->columns as $column) {
+
+                if ($column->defaultValue !== null) {
```
WinterSilence, line 350: `if ($column->defaultValue === null && !$column->allowNull) {
     continue;
}`

src/generators/model/Generator.php
uldisn: I know only MySql.

src/generators/model/Generator.php
uldisn: Not need "&& !$column->allowNull"

src/generators/model/Generator.php
uldisn: I'm sure just about MySql

src/generators/model/Generator.php
uldisn: ok

WinterSilence: @uldisn no need `$columnsDefaultNull` set `if ($column->allowNull) {$defaultValue = 'null';}`

uldisn: by default all attributes has null value. Not need.

src/generators/model/Generator.php
WinterSilence: @uldisn that's types normalized check DB subfolders, also:
~~~php
/**
 * @var string abstract type of this column. Possible abstract types include:
 * char, string, text, boolean, smallint, integer, bigint, float, decimal, datetime,
 * timestamp, time, date, binary, and money.
 */
public $type;
/**
 * @var string the PHP type of this column. Possible PHP types include:
 * `string`, `boolean`, `integer`, `double`, `array`.
 */
public $phpType;
~~~
as you can see, `$phpType` use only 5 variants - you detect integer as `if (in_array($column->phpType, ['integer', 'double', 'float'], true))` 'float' added because php dont have type 'double', i think it's used to set validation rule

src/generators/model/Generator.php
WinterSilence: @uldisn `null !== 'null'`

src/generators/model/Generator.php
WinterSilence: @uldisn yes, my fail
~~~
if (strtoupper($column->defaultValue) === 'CURRENT_TIMESTAMP') {
     if ($driverName === 'mysql' && $column->type === Schema::TYPE_TIMESTAMP) {
          $defaultValue = 'time()';
     } else {
        $defaultValue = "date('Y-m-d H:i:s')";
    }
}
~~~

src/generators/model/Generator.php
uldisn: I checked MySql schema. If in MySql field no default value, then $column->defaultValue is null.
$column->allowNull use for rules "required"
Correct:
```php
 if ($column->defaultValue === null) { continue; }
 ```

WinterSilence: @uldisn `$column->allowNull` is `@param type|null $foo`, `function setFoo(?type $value)` and `function getFoo(): ?type`.

>  If in MySql field no default value, then $column->defaultValue is null.

~~~sql
CREATE TABLE `demo` (`a` VARCHAR(255) NULL, `b` TEXT NULL);
INSERT INTO `demo` (`a`) VALUES (NULL); // (NULL, NULL)
~~~
+
~~~sql
CREATE TABLE ` demo` (
    `id` INT(11) UNSIGNED NOT NULL AUTO_INCREMENT,
    `a` TEXT NULL DEFAULT NULL,
    `b` VARCHAR(50) NULL DEFAULT 'is default'
    PRIMARY KEY (`id`) USING BTREE
) ENGINE=InnoDB;

INSERT INTO ` demo` (`a`, `b`) VALUES ('...', NULL);
INSERT INTO ` demo` (`a`, `b`) VALUES (NULL, '...');
INSERT INTO `demo` (`b`) VALUES (NULL);
INSERT INTO `demo` (`a`) VALUES ('...');
~~~

PR #421, opened 2019-11-20
Fix for issue #335

kjusupov:
Fix related to generation of model rules (setting default values) specific to Postgres (issue #335)

Note: I will update the original issue with more details.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | Not sure
| Fixed issues  | #335


src/generators/model/Generator.php
```diff
@@ -335,19 +337,24 @@ public function generateRules($table)
     {
         $types = [];
         $lengths = [];
+        $defValues = [];
```
samdark:
```suggestion
        $defaultValues = [];
```

Issue #422, opened 2019-11-22
Code generation codestyle issues


mikk150:
### What steps will reproduce the problem?
1. update yii2-gii to dev-master
2. generate models for database with some relations

### What's expected?
generated code should be beautiful. EG
```php
    /**
     * @return \yii\db\ActiveQuery|\common\models\queries\AuthItemQuery
     */
    public function getItemName()
    {
        return $this->hasOne(AuthItem::className(), ['name' => 'item_name']);
    }
```
### What do you get instead?
```php
    /**
     * @return \yii\db\ActiveQuery|\common\models\queries\AuthItemQuery     */
    public function getItemName()
    {
        return $this->hasOne(AuthItem::className(), ['name' => 'item_name']);
    }
```

### Additional info

| Q                | A
| ---------------- | ---
| Gii version      |  dev-master 2fdf98a
| PHP version      | 7.2.15
| Database version | 5.7
| Operating system | Docker

mikk150:
There is no `PHP_EOL` after `@return`

PR #423, opened 2019-12-10
fix relational query getter documentation style

mikk150:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #422


mikk150:
I could also add more comments if it is needed
Screenshot from 2019-12-10 16-33-57

samdark:
Yes. In this case extra doc would be useful.

samdark:
Merged. Thank you!

Issue #14, opened 2015-03-25
Using viaTable() or via() method


LAV45:
Why gii Model generator, by default, creates links between different models using the method viaTable() and not via()
```php
public function getPageTags()
{
    return $this->hasMany(PageTags::className(), ['page_id' => 'id']);
}

public function getTags()
{
    return $this->hasMany(Tag::className(), ['id' => 'tag_id'])
        //->viaTable('page_tag', ['page_id' => 'id']);
        ->via('pageTags');
}
```

if you change table name then it will have to fix in a few places

cebe:
in most cases you do not need a model for the junction table. Gii generates code for the most common case.

LAV45:
@samdark, what do you think about this?

I already have finished and tested the solution, I just need your approval.

samdark:
As @cebe said, using tables is a bit more common.

PR #424, opened 2019-12-15
Added support for via() relations in model generator

rhertogh:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #14 , #212

### DIFF
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -111,7 +111,7 @@ class Generator extends \yii\gii\Generator
             'modelClass' => 'Model Class Name',
             'baseClass' => 'Base Class',
             'generateRelations' => 'Generate Relations',
-            'generateJunctionRelationMode' => 'Generate Junction Relation Mode',
+            'generateJunctionRelationMode' => 'Generate Junction Relations As',
             'generateRelationsFromCurrentSchema' => 'Generate Relations from Current Schema',
             'generateLabelsFromComments' => 'Generate Labels from DB Comments',
             'generateQuery' => 'Generate ActiveQuery',
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -35,11 +35,6 @@ class Generator extends \yii\gii\Generator
     const JUNCTION_RELATION_VIA_TABLE = 'table';
     const JUNCTION_RELATION_VIA_MODEL = 'model';
 
-    const JUNCTION_RELATION_MODES = [
-        self::JUNCTION_RELATION_VIA_TABLE,
-        self::JUNCTION_RELATION_VIA_MODEL
-    ];
-
     public $db = 'db';
     public $ns = 'app\models';
     public $tableName;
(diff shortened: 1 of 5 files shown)


rhertogh:
Please note that that while the generated model is backwards compatible by default (viaTable is the default option), the signature of `Generator::checkJunctionTable()` has changed. This might cause problems when someone has extended the Generator and uses the checkJunctionTable function in custom code.
My suggestion is to add a `upgrade.md` to describe the change. Please let me know if you agree and I'll add it.

schmunk42:
@yiisoft/core-developers Please do a `2.2.x` release if a method signature is changed.

samdark:
@schmunk42 OK but I don't see any signature changes in the pull request.

rhertogh:
@samdark 
>  OK but I don't see any signature changes in the pull request.
My comment was a bit unclear, it's not the signature of the function but the signature of the returned value (the array now has an extra dimension).

rhertogh:
I've changed the changelog version in the pr  to 2.2.0 and added UPGRADE.md.
The Breaks BC? is changed to 'yes' in the pr opening post.

src/generators/model/Generator.php
```diff
@@ -104,6 +111,7 @@ public function attributeLabels()
             'modelClass' => 'Model Class Name',
             'baseClass' => 'Base Class',
             'generateRelations' => 'Generate Relations',
+            'generateJunctionRelationMode' => 'Generate Junction Relation Mode',
```
samdark:
```suggestion
            'generateJunctionRelationMode' => 'Generate Junction Relations As',
```

src/generators/model/Generator.php
```diff
@@ -490,25 +501,65 @@ private function generateManyManyRelations($table, $fks, $relations)
...
-            $relationName = $this->generateRelationName($relations, $table0Schema, key($secondKey), true);
-            $relations[$table0Schema->fullName][$relationName] = [
-                "return \$this->hasMany($className1::className(), $link)->viaTable('"
-                . $this->generateTableName($table->name) . "', $viaLink);",
-                $className1,
-                true,
-            ];
+            $link = $this->generateRelationLink(array_flip($secondKey[0]));
+            $relationName = $this->generateRelationName($relations, $table0Schema, key($secondKey[0]), true);
+            if ($this->generateJunctionRelationMode == self::JUNCTION_RELATION_VIA_TABLE) {
```
samdark:
```suggestion
            if ($this->generateJunctionRelationMode === self::JUNCTION_RELATION_VIA_TABLE) {
```

src/generators/model/Generator.php
```diff
@@ -490,25 +501,65 @@ private function generateManyManyRelations($table, $fks, $relations)
...
+                $relations[$table0Schema->fullName][$relationName] = [
+                    "return \$this->hasMany($className1::className(), $link)->viaTable('"
+                    . $this->generateTableName($table->name) . "', " . $this->generateRelationLink($firstKey[0]) . ");",
+                    $className1,
+                    true,
+                ];
+            } elseif ($this->generateJunctionRelationMode == self::JUNCTION_RELATION_VIA_MODEL) {
```
samdark:
```suggestion
            } elseif ($this->generateJunctionRelationMode === self::JUNCTION_RELATION_VIA_MODEL) {
```

src/generators/model/Generator.php
```diff
@@ -490,25 +501,65 @@ private function generateManyManyRelations($table, $fks, $relations)
...
+                    throw new Exception('Foreign key for junction table not found.');
+                }
+                $relations[$table0Schema->fullName][$relationName] = [
+                    "return \$this->hasMany($className1::className(), $link)->via('"
+                    . lcfirst($foreignRelationName) . "');",
+                    $className1,
+                    true,
+                ];
+            } else {
+                throw new InvalidConfigException('Unknown generateViaRelationMode ' . $this->generateJunctionRelationMode);
```
samdark: This check could be done once at the top of the method. Would be a bit less cognitive load.

rhertogh: Check is moved to the top of the function. Using new JUNCTION_RELATION_MODES array to check against. This array is now also used in the `rules()`.

src/generators/model/Generator.php
```diff
@@ -35,6 +35,11 @@ class Generator extends \yii\gii\Generator
     const JUNCTION_RELATION_VIA_TABLE = 'table';
     const JUNCTION_RELATION_VIA_MODEL = 'model';
 
+    const JUNCTION_RELATION_MODES = [
```
samdark: That won't work for older PHP versions.

rhertogh: dropped the array constant, now defining the array in functions where needed.

samdark:
Thank you!

PR #427, opened 2020-03-30
Update use of ::className() in model/Generator

yoshifp:
Update use of ::className() in model/Generator to ::class

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | N/A

### DIFF
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -86,8 +86,8 @@ class Generator extends \yii\gii\Generator
             [['ns', 'queryNs'], 'validateNamespace'],
             [['tableName'], 'validateTableName'],
             [['modelClass'], 'validateModelClass', 'skipOnEmpty' => false],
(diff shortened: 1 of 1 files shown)


samdark:
We support 5.4. Can't merge it.

rob006:
@samdark both app templates already requires PHP 5.6.

samdark:
Yes but framework is not app template.

rob006:
And Gii is not part of the core framework. App templates is just an example that we should move on - PHP 5.4 EOLed over 4 years ago. For me it a bug that Gii targets that old PHP version and generates code with methods that were deprecated in core.

samdark:
1. PHP version in composer.json should be adjusted.
2. https://travis-ci.com/github/yiisoft/yii2-gii/jobs/309318715

PR #428, opened 2020-04-03
Allow objects including anonymous classes on config

aguevaraIL:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

Example of what currently is not possible.

```php
    'modules' => [
        'gii' => [
            'class' => yii\gii\Module::class,
            'generators' => [
                'custom' => new class () extends Generator {
                    public function getName() {
                        return 'Custom Generator';
                    }
                    public function generate()
                    {
                       // ...
                    }
                },
            ],
        ],
```

It throws  exception

`Exception 'yii\base\InvalidConfigException' with message 'Unsupported configuration type: object'`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Yii Framework 2 gii extension Change Log
 2.2.1 under development
 -----------------------
 
-- Bug #428: Permit the usage of anonymous generators using dependency injection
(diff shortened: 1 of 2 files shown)


aguevaraIL:
@samdark done

samdark:
👍

Issue #439, opened 2021-02-28
Model generator ignore checkbox Singularize


WinterSilence:
> For example, table names like <code>some_tables</code> will have class names <code>SomeTable</code>.

### What steps will reproduce the problem?

1. Open page `gii/model` in browser
2. Check Singularize
3. Select table name, in my case it's "caterories"

### What's expected?

`Model Class Name` auto sets as `Caterories`. I check code - class name generated in JS and ignore Singularize

### What do you get instead?

Class name must be set manually by `yii\gii\generators\model\Generator::generateClassName()` as `Category`

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      |  2.0.40, gii 2.2.1
| PHP version      | 7.3
| Database version | msql8
| Operating system | Win10

samdark:
@WinterSilence any idea about the fix?

WinterSilence:
@samdark append ajax action returned generateClassName(). I'm install yii2 2 days ago :)

WinterSilence:
@samdark fix #447

WinterSilence:
@samdark are we waiting for?

samdark:
Code review. I'll do it soon. It's in my queue.

![image](https://user-images.githubusercontent.com/47294/113519698-93aaa800-9596-11eb-89b9-610a7c798987.png)

samdark:
Why closing? PR is still open, isn't it?

WinterSilence:
@samdark bug fixed, PR approved, nothing to discuss here

samdark:
@WinterSilence we usually keep the issue open until it's solved in master branch.

PR #443, opened 2021-03-24
Changed visibility of model generator methods

WinterSilence:
First step to fix issue #439

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes
| Fixed issues  | #439

### DIFF
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -992,10 +992,10 @@ class Generator extends \yii\gii\Generator
     /**
      * Generates a class name from the specified table name.
      * @param string $tableName the table name (which may contain schema prefix)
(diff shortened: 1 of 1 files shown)


bizley:
It would be better to make all changes in one PR. Please explain reasoning behind this.

WinterSilence:
@bizley I prefer to solve it step by step, small PR's = short discussion and quick adoption

WinterSilence:
@samdark do you have any other suggestions on how to fix  #439?

samdark:
@WinterSilence the question should be the opposite. Visibility of methods should stay as low as possible if they're not used from outside. From the PR it isn't clear what the change is for.

PR #447, opened 2021-03-30
Bugfix: generate model class in web app

WinterSilence:
Replace pseudo generation to request `gii/default/action` > `yii\gii\generators\model\Generator::actionGenerateClassName()`. 
Replace invalid  attribute(custom/user's attributes must have prefix `data-`)  `table_prefix` to `data-table-prefix`.

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #439

yii/gii use something like Selenium tests?

### DIFF
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -1046,8 +1046,8 @@ class Generator extends \yii\gii\Generator
 
     /**
      * Action to generate class name.
-     * 
      * @return string
+     * @since 2.2.2
      */
     public function actionGenerateClassName()
     {
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -1045,6 +1045,16 @@ class Generator extends \yii\gii\Generator
     }
 
     /**
+     * Action to generate class name.
+     * 
+     * @return string
+     */
+    public function actionGenerateClassName()
+    {
+        return $this->generateClassName($this->tableName);
+    }
+
+    /**
      * Generates a query class name from the specified model class name.
      * @param string $modelClassName model class name
      * @return string generated class name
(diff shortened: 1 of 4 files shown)


bizley:
I've tested and all works fine. I would just switch the order of inputs since these two checkboxes configure the output in model name field so I would have them above the `Model Class Name` text field.

image

src/generators/model/Generator.php
```diff
@@ -1044,6 +1044,16 @@ protected function generateClassName($tableName, $useSchemaName = null)
         return $this->classNames[$fullTableName];
     }
 
+    /**
+     * Action to generate class name.
+     * 
+     * @return string
```
bizley, line 1050:
```suggestion
     * @return string
     * @since 2.2.2
```

bizley (approved):

WinterSilence:
@bizley but then need move up 'db', 'useTablePrefix', 'useSchemaName' too. also, i find another bug - change 'db' don't have any effects on list of tables and relation fields

bizley:
Why? These don't affect model class name.

WinterSilence:
@bizley nope. If model field is empty then value generated by table field. if you move up model setting then you need table settings too.

bizley:
So yes, move it.

bizley:
You did. Great.

bizley:
Why is this closed?

WinterSilence:
@bizley :( my fail

bizley:
Ok, not a problem, it happens 😄

samdark:
Thanks!

PR #450, opened 2021-04-06
Update crud controller view

WinterSilence:
1. merge model behaviors with `parent::behaviors()`
2. replace `Yii::$app->request` to `$this->request`
3. fast check `$this->request->isPost` before load & save model
4. add commented example of pagination and sorting settings for `ActiveDataProvider`

| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/src/generators/crud/default/controller.php
+++ b/src/generators/crud/default/controller.php
@@ -70,7 +70,7 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
     {
 <?php if (!empty($generator->searchModelClass)): ?>
         $searchModel = new <?= isset($searchModelAlias) ? $searchModelAlias : $searchModelClass ?>();
-        $dataProvider = $searchModel->search($this->request->getQueryParams());
+        $dataProvider = $searchModel->search($this->request->queryParams);
 
         return $this->render('index', [
             'searchModel' => $searchModel,
--- a/src/generators/crud/default/controller.php
+++ b/src/generators/crud/default/controller.php
@@ -121,11 +121,13 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
     {
         $model = new <?= $modelClass ?>();
 
-        if ($this->request->isPost && $model->load($this->request->post()) && $model->save()) {
-            return $this->redirect(['view', <?= $urlParams ?>]);
+        if ($this->request->isPost) {
+            if ($model->load($this->request->post()) && $model->save()) {
+                return $this->redirect(['view', <?= $urlParams ?>]);
+            }
+        } else {
+            $model->loadDefaultValues();
         }
-
-        $model->loadDefaultValues();
 
         return $this->render('create', [
             'model' => $model,
--- a/src/generators/crud/default/controller.php
+++ b/src/generators/crud/default/controller.php
@@ -29,7 +29,6 @@ echo "<?php\n";
 
 namespace <?= StringHelper::dirname(ltrim($generator->controllerClass, '\\')) ?>;
 
-use Yii;
 use <?= ltrim($generator->modelClass, '\\') ?>;
 <?php if (!empty($generator->searchModelClass)): ?>
 use <?= ltrim($generator->searchModelClass, '\\') . (isset($searchModelAlias) ? " as $searchModelAlias" : "") ?>;
@@ -46,18 +45,21 @@ use yii\filters\VerbFilter;
 class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->baseControllerClass) . "\n" ?>
 {
     /**
-     * {@inheritdoc}
+     * @inheritDoc
      */
     public function behaviors()
     {
-        return [
-            'verbs' => [
-                'class' => VerbFilter::className(),
-                'actions' => [
-                    'delete' => ['POST'],
+        return array_merge(
+            parent::behaviors(),
+            [
+                'verbs' => [
+                    'class' => VerbFilter::className(),
+                    'actions' => [
+                        'delete' => ['POST'],
+                    ],
                 ],
-            ],
-        ];
+            ]
+        );
     }
 
     /**
@@ -68,7 +70,7 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
     {
 <?php if (!empty($generator->searchModelClass)): ?>
         $searchModel = new <?= isset($searchModelAlias) ? $searchModelAlias : $searchModelClass ?>();
-        $dataProvider = $searchModel->search(Yii::$app->request->queryParams);
+        $dataProvider = $searchModel->search($this->request->getQueryParams());
 
         return $this->render('index', [
             'searchModel' => $searchModel,
@@ -77,6 +79,18 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
 <?php else: ?>
         $dataProvider = new ActiveDataProvider([
             'query' => <?= $modelClass ?>::find(),
+            /*
+            'pagination' => [
+                'pageSize' => 50
+            ],
+            'sort' => [
+                'defaultOrder' => [
+<?php foreach ($pks as $pk): ?>
+                    <?= "'$pk' => SORT_DESC,\n" ?>
+<?php endforeach; ?>
+                ]
+            ],
+            */
         ]);
 
         return $this->render('index', [
@@ -107,9 +121,11 @@ class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->bas
     {
         $model = new <?= $modelClass ?>();
 
-        if ($model->load(Yii::$app->request->post()) && $model->save()) {
+        if ($this->request->isPost && $model->load($this->request->post()) && $model->save()) {
             return $this->redirect(['view', <?= $urlParams ?>]);
         }
+
+        $model->loadDefaultValues();
 
         return $this->render('create', [
             'model' => $model,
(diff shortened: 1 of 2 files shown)


src/generators/crud/default/controller.php
```diff
@@ -46,18 +45,21 @@
...
+     * @inheritDoc
      */
     public function behaviors()
     {
-        return [
-            'verbs' => [
-                'class' => VerbFilter::className(),
-                'actions' => [
-                    'delete' => ['POST'],
+        return array_merge(
```
bizley, line 52: What is the use case behind this?

src/generators/crud/default/controller.php
```diff
@@ -68,7 +70,7 @@ public function actionIndex()
     {
 <?php if (!empty($generator->searchModelClass)): ?>
         $searchModel = new <?= isset($searchModelAlias) ? $searchModelAlias : $searchModelClass ?>();
-        $dataProvider = $searchModel->search(Yii::$app->request->queryParams);
+        $dataProvider = $searchModel->search($this->request->getQueryParams());
```
bizley, line 73: That property exists since Yii 2.0.36, before that version user will get an error.

src/generators/crud/default/controller.php
WinterSilence: @bizley `baseControllerClass`:

~~~php
<?php

namespace app\controllers;

abstract class BackendController extends \yii\web\Controller
{
    public function behaviors()
    {
        return [
            'access' => [
                'class' => \yii\filters\AccessControl::class,
                'rules' => [[
                    'allow' => true,
                    'roles' => ['@'],
                ]],
            ],
        ];
    }
}
~~~

src/generators/crud/default/controller.php
WinterSilence: @bizley https://github.com/yiisoft/yii2-gii/blob/e11fec612df24f7d7223f133c23016003cae6987/composer.json#L26

src/generators/crud/default/controller.php
bizley: So you want to change the vendor default to suit your own implementation? `yii\web\Controller` does not contain behaviors config. I'm not sure if adding that merge (while not being wrong per se) is fine but I would like to hear other opinions.

src/generators/crud/default/controller.php
bizley: Yes, so anyone running Gii on Yii 2.0.14 to 2.0.35 will get an error - what is your point?

src/generators/crud/default/controller.php
WinterSilence: @bizley it's only example, current code ignore/remove `baseControllerClass::behaviors()`

src/generators/crud/default/controller.php
WinterSilence: @bizley don't understand you, `getQueryParams()`  exists in [2.0.14](https://github.com/yiisoft/yii2/blob/2.0.14/framework/web/Request.php#L625) - [2.0.35](https://github.com/yiisoft/yii2/blob/2.0.34/framework/web/Request.php#L653)

samdark: Then it's alright: https://github.com/yiisoft/yii2-gii/blob/master/composer.json#L26

src/generators/crud/default/controller.php
samdark: Current code doesn't extend from any base controller.

bizley: `Generator::$baseControllerClass` is `yii\web\Controller`. If you change it in the config, Gii is not checking new controller's code to make amendments.

WinterSilence: @bizley you can set **any** parent controller

part of `@app/config/web.php`:
~~~
  $config['bootstrap'][] = 'gii';
  $config['modules']['gii'] = [
      'class' => 'yii\gii\Module',
      'generators' => [
          'crud' => [
              'class' => 'app\gii\generators\crud\Generator',
              'modelClass' => 'app\\models\\',
              'enablePjax' => true,
              'baseControllerClass' => 'app\controllers\BackendController', // replace `yii\web\Controller`
              'viewPath' => '@app/views/backend/',
              'templates' => [
                  'app' => '@app/gii/generators/crud/app',
              ]
          ],
      ],
  ];
~~~

WinterSilence: @bizley hmm... why it's work in my app?
изображение

bizley: Yes, you can set any controller. That is the point of what I'm saying here. If in the base controller you chose, you throw exception in the `behaviors()` method, should Gii generate code not only merging `parent::behaviors()` but also using try/catch? Do you understand what I'm saying because we need to be on the same page here, otherwise the discussion is pointless.

WinterSilence: @bizley if user don't catch exception in `parent::behaviors()` then it's must be throwing in generated controllers too, it's better than erase user's code.

src/generators/crud/default/controller.php
```diff
@@ -77,6 +79,18 @@ public function actionIndex()
...
         $dataProvider = new ActiveDataProvider([
             'query' => <?= $modelClass ?>::find(),
+            /*
+            'pagination' => [
+                'pageSize' => 50
+            ],
+            'sort' => [
+                'defaultOrder' => [
+<?php foreach ($pks as $pk): ?>
+                    <?= "'$pk' => SORT_DESC,\n" ?>
```
samdark, line 89: Why DESC is default here? Just a template?

samdark (approved):
Looks alright overall. Please add a CHANGELOG.md line.

bizley (changes requested):

src/generators/crud/default/controller.php
WinterSilence: @samdark yes, just example/template

src/generators/crud/default/controller.php
```diff
@@ -107,10 +121,12 @@ public function actionCreate()
     {
         $model = new <?= $modelClass ?>();
 
-        if ($model->load(Yii::$app->request->post()) && $model->save()) {
+        if ($this->request->isPost && $model->load($this->request->post()) && $model->save()) {
             return $this->redirect(['view', <?= $urlParams ?>]);
         }
 
+        $model->loadDefaultValues();
```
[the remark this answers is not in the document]
WinterSilence: @bizley nope, we load defaults only for show form, if move up it's mixed with post data.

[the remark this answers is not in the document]
WinterSilence: @bizley if set `$this->searchModelClass = 'app\models\FooSearch'` by default, but in form you set property as '' (empty).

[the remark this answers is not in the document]
bizley: No, if we load defaults after the post fetch and there will be validation errors user should see them with the values that were there, not the default ones.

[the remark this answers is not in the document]
WinterSilence: @bizley you right but not in all, it's must be:
~~~php
if ($this->request->isPost) {
    if ($model->load($this->request->post()) && $model->save()) {
    }
} else {
   $model->loadDefaultValues();
}
~~~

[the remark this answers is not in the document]
bizley: It will be enough if you move it up as I said. Why you don't want to show user the default values already prefilled?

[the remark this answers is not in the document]
WinterSilence: @bizley  I am not sure that in this case the empty values passed from the form will not be overwritten.
> if set $this->searchModelClass = 'app\models\FooSearch' by default, but in form you set property as '' (empty).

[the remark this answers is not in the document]
bizley: Sending empty value overwrites the default one (as long as there is validation rule for that field so `load()` can pick it up) so it's safe to leave it like I asked to.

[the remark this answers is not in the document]
WinterSilence: @bizley can you explain: why my version worst than your?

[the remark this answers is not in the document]
bizley: Unnecessary `else`.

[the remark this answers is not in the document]
WinterSilence: @bizley nope, I want pass defaults only on form view and skip it on submit because it's unuseful - defaults rewrites by form data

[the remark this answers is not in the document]
bizley: It does not, already told you.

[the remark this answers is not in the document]
WinterSilence: @bizley now, `$model->loadDefaultValues();` calls only if it's **not POST** request.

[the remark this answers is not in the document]
bizley: Man, I don't understand you. You add PRs like #451 removing unnecessary variables but it looks like you will to fight to death for unnecessary `else` here. Sigh...

samdark:
Thanks!

src/generators/crud/default/controller.php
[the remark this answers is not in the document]
WinterSilence: @bizley if you don't understand why it's added, then just check how it work on real model. i do it and it's work without any problems.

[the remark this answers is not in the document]
WinterSilence: @bizley you version work longer because call `$model->loadDefaultValues();` in any case

PR #455, opened 2021-05-02
Update main.scss

WinterSilence:
Add styles for preview sources

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | no

### DIFF
--- a/src/assets/scss/main.scss
+++ b/src/assets/scss/main.scss
@@ -371,3 +371,13 @@ body {
     height: 1px;
     padding: 0;
 }
+
+#preview-modal {
+    .modal-header {
+        display: flex;
+    }
+    .modal-body {
+        white-space: pre;
+        overflow: auto;
+    }
+}


samdark:
Would you please post before/after screenshots? What does it fix?

WinterSilence:
@samdark for example, check `composer.json`  generated by "Generate extension". By default, all prepended whitespaces are combined.

rob006:
IMO this should be handled in the same way as for PHP code: wrap content in `<code>` tag and replace spaces by `&nbsp;`.

WinterSilence:
@rob006 it's too good idea, but we don't have function/method to do this.

rob006:
To do what?

WinterSilence:
@rob006 to

> wrap content in <code> tag and replace spaces by &nbsp;

need replace only initial tabs\spaces in lines and it's give same result (`white-space: pre;`)

rob006:
You can just replace https://github.com/yiisoft/yii2-gii/blob/6b40893f87c1dd970896ced6c198ff2416df353e/src/CodeFile.php#L153 to something like 

```php
return strtr(nl2br(Html::encode($this->content)), [
    ' ' => 'nbsp;',
    "\t" => 'nbsp;nbsp;nbsp;nbsp;',
]);
```

At least you're sure nothing else is affected by this change (your PR affects all modals - are you sure modals contain only code?) and style is consistent with PHP code (currently proportional font is used for non-PHP code snippets).

WinterSilence:
@rob006
> your PR affects all modals - are you sure modals contain only code?)

nope, all modals - `.modal`, i change styles only for `#preview-modal`

PR #458, opened 2021-07-17
Add CIDR support for allowedIPs

rhertogh:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

PR #459, opened 2021-07-17
Added unit test for Module::checkAccess()

rhertogh:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes (but requires #458)
| Fixed issues  | -

### DIFF
--- a/tests/ModuleTest.php
+++ b/tests/ModuleTest.php
@@ -19,4 +19,83 @@ class ModuleTest extends TestCase
 
         $this->assertEquals('2.0.6', $module->getVersion());
     }
+
+    /**
+     * Data provider for [[testCheckAccess()]]
+     * @return array test data
+     */
+    public function dataProviderCheckAccess()
+    {
+        return [
+            [
+                [],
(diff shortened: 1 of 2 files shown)


rhertogh:
Requires #458 for `testCheckAccess` to pass.

rob006:
If these are tests for #458, they should be added to #458. Creating separate PR for tests only is confusing and harder to review.

rhertogh:
> If these are tests for #458, they should be added to #458. Creating separate PR for tests only is confusing and harder to review.

These tests are not specific for #458 but mainly for existing code, only a small portion is covering 458. I could remove that portion from this pull request, but then non-related code would be added to 458 or a third pull request would be required.

rob006:
I don't see a problem to include these tests in #458, they are related and it certainly makes more sense than dependency on PR without tests (so you need to merge PR without knowing it actually works).

bizley:
Please add those tests to #458

PR #461, opened 2021-08-05
Fix #435: Fix incorrect classNames

### LINKED ISSUE #435: gii/model --tableName=prefix_* generates classnames from the full table name instead of from the matching characters as indicated by --help
1. Make sure the default `db` Connection does not have `tablePrefix` set.
2. Create two or more tables with some prefix, for example: `prefix_some_model`, `prefix_another_model`
3. run the command `./yii gii/model --tableName=prefix_*`

### What's expected?
Create two classes, named: SomeModel and AnotherModel

### What do you get instead?
Two classes, named PrefixSomeModel and PrefixAnotherModel

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      |  2.0.34
| Yii Gii version      |  2.1.4
| PHP version      | 7.3.14
| Database version | MariaDB v15.1
| Operating system | Debian 10.3

`yii\gii\generators\model\Generator::generateClassName()` generates classnames by matching the following patterns:
```php
        $patterns[] = "/^{$db->tablePrefix}(.*?)$/";
        $patterns[] = "/^(.*?){$db->tablePrefix}$/";
        if (strpos($this->tableName, '*') !== false) {
            $pattern = $this->tableName;
            if (($pos = strrpos($pattern, '.')) !== false) {
                $pattern = substr($pattern, $pos + 1);
            }
            $patterns[] = '/^' . str_replace('*', '(\w+)', $pattern) . '$/';
        }
```
When $db->tablePrefix is empty, this results in an array containing three patterns of which the first two would both be `/^(.*?)$/` and the third `/^prefix_(\w+)$/`. Since patterns are matched in the order in which they appear in the array, the prefix pattern would never match. 

Solution would be to check if `$db->tablePrefix` is empty and not set or change the order of the patterns.

ditibal:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #435

### DIFF
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -1012,8 +1012,6 @@ class Generator extends \yii\gii\Generator
 
         $db = $this->getDbConnection();
         $patterns = [];
-        $patterns[] = "/^{$db->tablePrefix}(.*?)$/";
-        $patterns[] = "/^(.*?){$db->tablePrefix}$/";
         if (strpos($this->tableName, '*') !== false) {
             $pattern = $this->tableName;
             if (($pos = strrpos($pattern, '.')) !== false) {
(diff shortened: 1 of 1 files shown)


WinterSilence:
hmm.. you move rules, what's changed?

samdark:
@WinterSilence see issue linked.

WinterSilence:
@samdark it's need fix, but not use other way: current solution BC, need add option to switch behavior

PR #462, opened 2021-08-09
Added support for viewing file difference on CLI

rhertogh:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | -

### DIFF
--- a/src/console/GenerateAction.php
+++ b/src/console/GenerateAction.php
@@ -72,12 +72,21 @@ class GenerateAction extends \yii\base\Action
                     if ($skipAll !== null) {
                         $answers[$file->id] = !$skipAll;
                     } else {
-                        $answer = $this->controller->select("Do you want to overwrite this file?", [
-                            'y' => 'Overwrite this file.',
-                            'n' => 'Skip this file.',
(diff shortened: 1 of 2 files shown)


samdark (approved):

src/console/GenerateAction.php
```diff
@@ -72,12 +72,21 @@ protected function generateCode()
...
+                            $answer = $this->controller->select("Do you want to overwrite this file?", [
+                                'y' => 'Overwrite this file.',
+                                'n' => 'Skip this file.',
+                                'ya' => 'Overwrite this and the rest of the changed files.',
+                                'na' => 'Skip this and the rest of the changed files.',
+                                'v' => 'View difference',
+                            ]);
+
+                            if ($answer === 'v') {
+                                $diff = new \Diff(explode("\n", file_get_contents($file->path)), explode("\n", $file->content));
```
bizley, line 85: It looks like it's worth to extract `file_get_contents($file->path)` to a variable, especially that it's called in a loop but also it is already called at line 65.

samdark:
Thanks!

PR #453, opened 2021-04-15
Allow CRUD to work with non-RDBMS ARs

WinterSilence:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #452

PR #466, opened 2021-09-24
Fix #463: added back special case (single PK)

### LINKED ISSUE #463: Gii only works for tables with ID as a PK?
I am using https://www.yiiframework.com/doc/guide/2.0/en/start-databases as a starting point. 

It recommends to create a table with the following structure 
CREATE TABLE `country` (
  `code` CHAR(2) NOT NULL PRIMARY KEY,
  `name` CHAR(52) NOT NULL,
  `population` INT(11) NOT NULL DEFAULT '0'
) ENGINE=InnoDB DEFAULT CHARSET=utf8;

I can then create a model and the CRUD methods using the gii generator.

The issue is that the generated code refers to id as a primary key at all times, not the case due to the suggested table structure, and even the findModel needs to be updated from.

```
protected function findModel($code)
 {
    if (($model = Country2::findOne($id)) !== null) {
        return $model;
    }

    throw new NotFoundHttpException('The requested page does not exist.');
} 
```
To
```

 protected function findModel($code)
 {
    if (($model = Country2::findOne($code)) !== null) {
        return $model;
    }

    throw new NotFoundHttpException('The requested page does not exist.');
}
```

This makes everything work once you create a Country and you can edit it, delete it... 

The issue now is with the basic gridview on index where every action has an id tacked on to it.

ie https://localhost/tutorialYii/web/index.php?r=country2%2Fview&id=12 rather than https://localhost/tutorialYii/web/index.php?r=country2%2Fview&code=12

I understand this can be fixed by using id as the primary key,done so to test it worked, but I would like to understand how to address the issue of using a custom primary key rather than the standard one.

I know the workaround of using id as a PK, I would just like to understand how to get a gridview that works as intended with a different PK.

cjrf:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  | #463

### DIFF
--- a/src/generators/crud/Generator.php
+++ b/src/generators/crud/Generator.php
@@ -495,6 +495,14 @@ class Generator extends \yii\gii\Generator
         $class = $this->modelClass;
         $pks = $class::primaryKey();
         $isMongoModel = is_subclass_of($class, '\yii\mongodb\ActiveRecord');
+        if (count($pks) === 1) {
+            if ($isMongoModel) {
+                return "'id' => (string)\$model->{$pks[0]}";
+            }
+
+            return "'id' => \$model->{$pks[0]}";
+        }
+
         $params = [];
         foreach ($pks as $pk) {
             if ($isMongoModel) {
@@ -516,6 +524,10 @@ class Generator extends \yii\gii\Generator
     {
         $class = $this->modelClass;
         $pks = $class::primaryKey();
+        if (count($pks) === 1) {
+            return '$id';
+        }
+
         return '$' . implode(', $', $pks);
     }
 


rob006:
Wouldn't be better to use actual param name [here](https://github.com/yiisoft/yii2-gii/blob/622761f492f46fb0a18bd026bcc32c243e642ce8/src/generators/crud/default/controller.php#L182)?

WinterSilence:
Better fix https://github.com/yiisoft/yii2-gii/pull/467

samdark:
Fixed another way as @WinterSilence suggested. Please check if it works now.

cjrf:
Still breaks grid/ActionColumn.

WinterSilence:
@cjrf can you post more details?

cjrf:
For years now, there have been four snippets of code that "rename a single PK to id". So I assume this is a feature. 

Three of those snippets where in Gii, one in yii/grid/ActionColumn::createUrl. 

Fix #453 removed just 2/3 of the Gii snippets, breaking the feature and Gii itself. 

My suggestion (#466) was to simply put those snippets back. 

Fix #467 removed the third Gii snippet, fixing Gii itself, but not the feature. 

> However, grid/ActionColumn probably shouldn't force "id" as the PK name.

By this, I mean that one solution could be to change yii/grid/ActiveColumn and remove completely this renaming feature from Yii2. I advise against this, since it will break back compatibility.

WinterSilence:
@cjrf ok, now I'm understand you, but rollback does not solve current problem - primary key can consist of several columns. We need further proceedings. If you have any ideas how fix it without BC, then please post it in new issue.

rob006:
@cjrf We can generate `urlCreator` by Gii and overwrite default `ActionColumn::createUrl()` - in that case old code will work as before, and new templates will use new param schema everywhere. 

I'm still not sure if it is worth the trouble. It should be much simpler to just revert #467 and apply this PR instead. Also, using `id` as param name may be more practical from `UrlManager` perspective - you can use single standard `<controller>/<action>/<id>` rule for all CRUDs, even if you name your PK as `post_id`, `thread_id`, etc.

@WinterSilence I don't think that any of recent changes affected composite private keys.

WinterSilence:
@rob006 
> We can generate urlCreator by Gii and overwrite default ActionColumn::createUrl()

I suggested same thing to @samdark

WinterSilence:
@rob006
> I'm still not sure if it is worth the trouble. It should be much simpler to just revert #467 and apply this PR instead

As I'm already said
> but rollback does not solve current problem - primary key can consist of several columns

rob006:
@WinterSilence Did you actually tested that this problem exists if you revert #467 and apply #466? Or this is just your speculations?

WinterSilence:
@rob006 https://github.com/yiisoft/yii2/blob/1d7f9d8d2da7ae3497d7c18b215b1cb569305ee2/framework/grid/ActionColumn.php#L222 ActionColumn not support it

rob006:
For me `is_array($key)` check indicates the opposite. But I also do not get how does it matter in terms of #467 and #466 - both PRs affects only simple PK and do not change behavior for composite PK.

WinterSilence:
@rob006 wrong link, I mean https://github.com/yiisoft/yii2-gii/blob/master/src/generators/crud/default/views/index.php#L36 
https://github.com/yiisoft/yii2/blob/1d7f9d8d2da7ae3497d7c18b215b1cb569305ee2/framework/grid/ActionColumn.php#L222 is worst solution in any case because we don't know how pk column(s) called in route, need bind route placeholder(s) to pk column(s)

cjrf:
> rollback does not solve current problem - primary key can consist of several columns
That problem doesn't exist. The current implementation of ActionColumn does support composite primary keys. In that case, no renaming takes place. 

> For me `is_array($key)` check indicates the opposite.
And you're correct. I wasn't sure but tested it and confirmed ActionColumn does support composite primary keys as it is. 

> using `id` as param name may be more practical from `UrlManager` perspective - you can use single standard `<controller>/<action>/<id>` rule for all CRUDs, even if you name your PK as `post_id`, `thread_id`, etc.

Exactly. That's probably why this renaming feature was added a long time ago.

WinterSilence:
@rob006 at least something in this world remains unchanged (:

PR #469, opened 2021-09-25
Update view `generators/crud/default/controller`

WinterSilence:
Fix return type in PhpDoc
Add Php 7 features (optional)
Use `Yii::createObject()` instead `new className()`

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/src/generators/crud/default/controller.php
+++ b/src/generators/crud/default/controller.php
@@ -1,125 +1,135 @@
 <?php
 /**
  * This is the template for generating a CRUD controller class file.
+ *
+ * @var \yii\web\View $this
+ * @var \yii\gii\generators\crud\Generator $generator
  */
 
 use yii\db\ActiveRecordInterface;
 use yii\helpers\StringHelper;
 
-
-/* @var $this yii\web\View */
-/* @var $generator yii\gii\generators\crud\Generator */
-
 $controllerClass = StringHelper::basename($generator->controllerClass);
 $modelClass = StringHelper::basename($generator->modelClass);
-$searchModelClass = StringHelper::basename($generator->searchModelClass);
-if ($modelClass === $searchModelClass) {
-    $searchModelAlias = $searchModelClass . 'Search';
+if (!empty($generator->searchModelClass)) {
+    $searchModelClass = StringHelper::basename($generator->searchModelClass);
+    if ($modelClass === $searchModelClass) {
+        $searchModelAlias = $searchModelClass . 'Search';
+    }
 }
 
-/* @var $class ActiveRecordInterface */
-$class = $generator->modelClass;
+/** @var $class ActiveRecord */
+$class = ltrim($generator->modelClass, '\\');
 $pks = $class::primaryKey();
 $urlParams = $generator->generateUrlParams();
 $actionParams = $generator->generateActionParams();
 $actionParamComments = $generator->generateActionParamComments();
+$php7 = PHP_MAJOR_VERSION === 7;
 
 echo "<?php\n";
 ?>
 
 namespace <?= StringHelper::dirname(ltrim($generator->controllerClass, '\\')) ?>;
 
-use <?= ltrim($generator->modelClass, '\\') ?>;
+use <?= $class ?>;
 <?php if (!empty($generator->searchModelClass)): ?>
-use <?= ltrim($generator->searchModelClass, '\\') . (isset($searchModelAlias) ? " as $searchModelAlias" : "") ?>;
+use <?= ltrim($generator->searchModelClass, '\\') . (isset($searchModelAlias) ? " as $searchModelAlias" : '') ?>;
 <?php else: ?>
 use yii\data\ActiveDataProvider;
 <?php endif; ?>
 use <?= ltrim($generator->baseControllerClass, '\\') ?>;
+use yii\web\Response;
 use yii\web\NotFoundHttpException;
 use yii\filters\VerbFilter;
+use Yii;
 
 /**
- * <?= $controllerClass ?> implements the CRUD actions for <?= $modelClass ?> model.
+ * Controller implements the CRUD actions for `<?= $modelClass ?>` model.
  */
 class <?= $controllerClass ?> extends <?= StringHelper::basename($generator->baseControllerClass) . "\n" ?>
 {
     /**
      * @inheritDoc
      */
-    public function behaviors()
+    public function behaviors()<?= ($php7 ? ': array' : '') . "\n" ?>
     {
         return array_merge(
             parent::behaviors(),
             [
-                'verbs' => [
-                    'class' => VerbFilter::className(),
+                'verbFilter' => [
+                    'class' => VerbFilter::class<?= $php7 ? '' : 'Name()' ?>,
                     'actions' => [
+                        'index' => ['GET'],
+                        'view' => ['GET'],
+                        'create' => ['GET', 'POST'],
+                        'update' => ['GET', 'POST'],
                         'delete' => ['POST'],
-                    ],
-                ],
+                    ]
+                ]
             ]
         );
     }
 
     /**
      * Lists all <?= $modelClass ?> models.
-     * @return mixed
+     * @return string
      */
-    public function actionIndex()
+    public function actionIndex()<?= ($php7 ? ': string' : '') . "\n" ?>
     {
 <?php if (!empty($generator->searchModelClass)): ?>
-        $searchModel = new <?= isset($searchModelAlias) ? $searchModelAlias : $searchModelClass ?>();
+        $searchModel = Yii::createObject(<?= isset($searchModelAlias) ? $searchModelAlias : $searchModelClass ?>::class<?= $php7 ? '' : 'Name()' ?>);
         $dataProvider = $searchModel->search($this->request->queryParams);
 
-        return $this->render('index', [
-            'searchModel' => $searchModel,
-            'dataProvider' => $dataProvider,
-        ]);
+        return $this->render(
+            'index',
+            [
+                'searchModel' => $searchModel,
+                'dataProvider' => $dataProvider,
+            ]
+        );
 <?php else: ?>
-        $dataProvider = new ActiveDataProvider([
-            'query' => <?= $modelClass ?>::find(),
-            /*
-            'pagination' => [
-                'pageSize' => 50
-            ],
-            'sort' => [
-                'defaultOrder' => [
+        $dataProvider = Yii::createObject(
+            ActiveDataProvider::class<?= $php7 ? '' : 'Name()' ?>,
+            [
+                'query' => <?= $modelClass ?>::find(),
+                /*
+                'pagination' => [
+                    'pageSize' => 50
+                ],
+                'sort' => [
+                    'defaultOrder' => [
 <?php foreach ($pks as $pk): ?>
-                    <?= "'$pk' => SORT_DESC,\n" ?>
+                        <?= "'$pk' => SORT_DESC,\n" ?>
 <?php endforeach; ?>
+                    ]
                 ]
-            ],
-            */
-        ]);
+                */
+            ]
+        );
 
-        return $this->render('index', [
-            'dataProvider' => $dataProvider,
-        ]);
+        return $this->render('index', ['dataProvider' => $dataProvider]);
 <?php endif; ?>
     }
 
     /**
      * Displays a single <?= $modelClass ?> model.
      * <?= implode("\n     * ", $actionParamComments) . "\n" ?>
-     * @return mixed
+     * @return string
      * @throws NotFoundHttpException if the model cannot be found
      */
-    public function actionView(<?= $actionParams ?>)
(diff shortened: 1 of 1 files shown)


src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
      * Lists all <?= $modelClass ?> models.
-     * @return mixed
+     * @return string
      */
-    public function actionIndex()
+    public function actionIndex()<?= ($php7 ? ': string' : '') . "\n" ?>
     {
 <?php if (!empty($generator->searchModelClass)): ?>
-        $searchModel = new <?= isset($searchModelAlias) ? $searchModelAlias : $searchModelClass ?>();
+        $searchModel = Yii::createObject(<?= isset($searchModelAlias) ? $searchModelAlias : $searchModelClass ?>::class<?= $php7 ? '' : 'Name()' ?>);
```
samdark, line 80: There is no need for `Yii::createObject` in this case since there's no benefit in storing search model in DI container.

samdark: PHP 7 `::class` is alright 👍

src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
-
-/* @var $this yii\web\View */
-/* @var $generator yii\gii\generators\crud\Generator */
-
 $controllerClass = StringHelper::basename($generator->controllerClass);
 $modelClass = StringHelper::basename($generator->modelClass);
-$searchModelClass = StringHelper::basename($generator->searchModelClass);
-if ($modelClass === $searchModelClass) {
-    $searchModelAlias = $searchModelClass . 'Search';
+if (!empty($generator->searchModelClass)) {
```
samdark, line 14: Why do we need this check?

src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
+    $searchModelClass = StringHelper::basename($generator->searchModelClass);
+    if ($modelClass === $searchModelClass) {
+        $searchModelAlias = $searchModelClass . 'Search';
+    }
 }
 
-/* @var $class ActiveRecordInterface */
-$class = $generator->modelClass;
+/** @var $class ActiveRecord */
```
samdark, line 21: Why changing it to concreate implementation?

src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
+$class = ltrim($generator->modelClass, '\\');
 $pks = $class::primaryKey();
 $urlParams = $generator->generateUrlParams();
 $actionParams = $generator->generateActionParams();
 $actionParamComments = $generator->generateActionParamComments();
+$php7 = PHP_MAJOR_VERSION === 7;
```
samdark, line 27:
```suggestion
$isPhp7 = PHP_MAJOR_VERSION === 7;
```

src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
 <?php endif; ?>
 use <?= ltrim($generator->baseControllerClass, '\\') ?>;
+use yii\web\Response;
 use yii\web\NotFoundHttpException;
 use yii\filters\VerbFilter;
+use Yii;
 
 /**
- * <?= $controllerClass ?> implements the CRUD actions for <?= $modelClass ?> model.
+ * Controller implements the CRUD actions for `<?= $modelClass ?>` model.
```
samdark, line 47: Why removing `<?= $controllerClass ?>`?

src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
      */
-    public function behaviors()
+    public function behaviors()<?= ($php7 ? ': array' : '') . "\n" ?>
     {
         return array_merge(
             parent::behaviors(),
             [
-                'verbs' => [
-                    'class' => VerbFilter::className(),
+                'verbFilter' => [
```
samdark, line 59: Why renaming the key?

src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
 <?php if (!empty($generator->searchModelClass)): ?>
-        $searchModel = new <?= isset($searchModelAlias) ? $searchModelAlias : $searchModelClass ?>();
+        $searchModel = Yii::createObject(<?= isset($searchModelAlias) ? $searchModelAlias : $searchModelClass ?>::class<?= $php7 ? '' : 'Name()' ?>);
         $dataProvider = $searchModel->search($this->request->queryParams);
 
-        return $this->render('index', [
-            'searchModel' => $searchModel,
-            'dataProvider' => $dataProvider,
-        ]);
+        return $this->render(
```
samdark, line 83: Previous formatting was OK.

src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
 <?php else: ?>
-        $dataProvider = new ActiveDataProvider([
-            'query' => <?= $modelClass ?>::find(),
-            /*
-            'pagination' => [
-                'pageSize' => 50
-            ],
-            'sort' => [
-                'defaultOrder' => [
+        $dataProvider = Yii::createObject(
```
samdark, line 91: There's no need for `Yii::createObject` here since there's no sense in putting `ActiveDataProvider` into container. It is not a service.

src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
-            */
-        ]);
+                */
+            ]
+        );
 
-        return $this->render('index', [
-            'dataProvider' => $dataProvider,
-        ]);
+        return $this->render('index', ['dataProvider' => $dataProvider]);
```
samdark, line 110: Formatting was alright.

src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
+     * @return string
      * @throws NotFoundHttpException if the model cannot be found
      */
-    public function actionView(<?= $actionParams ?>)
+    public function actionView(<?= $actionParams ?>)<?= ($php7 ? ': string' : '') . "\n" ?>
     {
-        return $this->render('view', [
-            'model' => $this->findModel(<?= $actionParams ?>),
-        ]);
+        return $this->render('view', ['model' => $this->findModel(<?= $actionParams ?>)]);
```
samdark, line 122: Formatting was alright.

src/generators/crud/default/controller.php
```diff
@@ -1,125 +1,135 @@
...
     /**
      * Creates a new <?= $modelClass ?> model.
      * If creation is successful, the browser will be redirected to the 'view' page.
-     * @return mixed
+     * @return Response|string
      */
     public function actionCreate()
     {
-        $model = new <?= $modelClass ?>();
+        $model = Yii::createObject(<?= $modelClass ?>);
```
samdark, line 132: No need to use `Yii::createObject` here.

src/generators/crud/default/controller.php
```diff
@@ -129,16 +139,14 @@ public function actionCreate()
             $model->loadDefaultValues();
         }
 
-        return $this->render('create', [
-            'model' => $model,
-        ]);
+        return $this->render('create', ['model' => $model]);
```
samdark, line 142: Formatting was OK.

src/generators/crud/default/controller.php
```diff
@@ -149,19 +157,17 @@ public function actionUpdate(<?= $actionParams ?>)
             return $this->redirect(['view', <?= $urlParams ?>]);
         }
 
-        return $this->render('update', [
-            'model' => $model,
-        ]);
+        return $this->render('update', ['model' => $model]);
```
samdark, line 160: Formatting was OK.

src/generators/crud/default/controller.php
```diff
@@ -149,19 +157,17 @@ public function actionUpdate(<?= $actionParams ?>)
...
     /**
      * Deletes an existing <?= $modelClass ?> model.
      * If deletion is successful, the browser will be redirected to the 'index' page.
      * <?= implode("\n     * ", $actionParamComments) . "\n" ?>
-     * @return mixed
+     * @return Response
      * @throws NotFoundHttpException if the model cannot be found
      */
-    public function actionDelete(<?= $actionParams ?>)
+    public function actionDelete(<?= $actionParams ?>): Response
```
samdark, line 170: Would error in lower-edge PHP versions supported.

src/generators/crud/default/controller.php
```diff
@@ -175,7 +181,7 @@ public function actionDelete(<?= $actionParams ?>)
      * @return <?=                   $modelClass ?> the loaded model
      * @throws NotFoundHttpException if the model cannot be found
      */
-    protected function findModel(<?= $actionParams ?>)
+    protected function findModel(<?= $actionParams ?>): <?= $modelClass . "\n" ?>
```
samdark, line 184: Would error in lower-edge PHP versions supported.

src/generators/crud/default/controller.php
```diff
@@ -184,10 +190,11 @@ protected function findModel(<?= $actionParams ?>)
 }
 $condition = '[' . implode(', ', $condition) . ']';
 ?>
-        if (($model = <?= $modelClass ?>::findOne(<?= $condition ?>)) !== null) {
```
samdark, line 187: It was OK as is. No need for change here.

samdark (changes requested):

src/generators/crud/default/controller.php
WinterSilence: @samdark because `$searchModelClass` not required

src/generators/crud/default/controller.php
WinterSilence: @samdark my mistake =\

src/generators/crud/default/controller.php
samdark: It will submit `''` anyway from the HTML form.

src/generators/crud/default/controller.php
WinterSilence: @samdark it's comment to this controller - class name is not informative in current case

src/generators/crud/default/controller.php
WinterSilence: @samdark I constantly forget that min version less 5.6

src/generators/crud/default/controller.php
WinterSilence: @samdark it's less readable

src/generators/crud/default/controller.php
WinterSilence: @samdark because in `AccessControl` too have `verbs` + key = class - easy rule doing code more readable

src/generators/crud/default/controller.php
WinterSilence: @samdark I reinsure in `null` case

src/generators/crud/default/controller.php
samdark: I think it was alright as is. No need to change it.

src/generators/crud/default/controller.php
samdark: How's `AccessControl` related? We're not using it here.

src/generators/crud/default/controller.php
samdark: OK.

src/generators/crud/default/controller.php
samdark: No, it's not. That's more of personal preferences. I'd not change it for that reason.

src/generators/crud/default/controller.php
WinterSilence: related via framework

src/generators/crud/default/controller.php
WinterSilence: psr-12-compatible style. yes, yii uses prev psr style, but we generate code for users

src/generators/crud/default/controller.php
WinterSilence: @samdark https://www.javaer101.com/en/article/2637550.html

src/generators/crud/default/controller.php
samdark: What do you mean? That's the name of the behavior registered. There's no another behavior so a longer name doesn't make sense to me.

src/generators/crud/default/controller.php
samdark: It was PSR-12 compatible.

src/generators/crud/default/controller.php
samdark: No need for `Yii::createObject` here. Search model is not a service. There are no defaults to configure.

src/generators/crud/default/controller.php
samdark: Other than that, you've changed early return to late return complicating the code a bit. With this assign in condition I agree.

Issue #437, opened 2020-11-03
className()  appear method is deprecated


haohetao:
### What steps will reproduce the problem?
Much of the code generated by gii have className() method.
### What's expected?
![图片](https://user-images.githubusercontent.com/4592571/98029737-f4e51880-1e4a-11eb-8116-de7287c5bb7e.png)

### What do you get instead?
using ::class insted of 

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.38
| PHP version      | 7.4,8.0
| Database version | pgsql
| Operating system | none

bizley:
It's because Yii 2 still supports PHP 5.4 and `::class` has been introduced in 5.5.

rob006:
> It's because Yii 2 still supports PHP 5.4 and `::class` has been introduced in 5.5.

It does not mean that Gii needs to generate code for PHP 5.4. Several Yii2 packages increased their PHP requirements, Gii could do the same: https://github.com/yiisoft/yii2-gii/issues/356#issuecomment-504798590

bizley:
I'm all for it but would be nice to do it for all dependent packages.

rob006:
App templates already require PHP 5.6, queue extension requires PHP 5.5. Are there any other official extensions that depends on Gii?

bizley:
@samdark what do you think? Are we upgrading dependants? And to which PHP version?

haohetao:
No one is using php5.4 now

samdark:
@bizley it is OK to upgrade apps since new apps are usually started with new PHP versions but I'm not so sure about Gii generating PHP 5.5+ syntax. From my experience, there are always projects with ancient PHP versions on the server that are, nevertheless, being in constant development.

rob006:
@samdark They can always use old Gii targeting PHP 5.4 syntax. There is no point in making this package less convenient for 99% of users only because someone stuck on PHP 5.4.

samdark:
If that's configurable then I'm all for it.

bizley:
@samdark I'm all for upgrade, If this would be possible I would upgrade to 7.4 ;)
@haohetao you would be surprised how many people are using old unsupported versions of PHP...

PR #470, opened 2021-09-25
Update index.php

WinterSilence:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |

### DIFF
--- a/src/generators/crud/default/views/index.php
+++ b/src/generators/crud/default/views/index.php
@@ -1,79 +1,93 @@
 <?php
 
+/**
+ * @var \yii\web\View $this
+ * @var \yii\gii\generators\crud\Generator $generator
+ */
+
 use yii\helpers\Inflector;
 use yii\helpers\StringHelper;
 
-/* @var $this yii\web\View */
-/* @var $generator yii\gii\generators\crud\Generator */
+$php7 = PHP_MAJOR_VERSION === 7;
+$modelClass = StringHelper::basename($generator->modelClass);
 
 echo "<?php\n";
 ?>
 
+/**
+ * @var \yii\web\View $this
+<?= !empty($generator->searchModelClass) ? " * @var " . $generator->searchModelClass . " \$searchModel\n" : '' ?>
+ * @var \yii\data\ActiveDataProvider $dataProvider
+ */
+
 use yii\helpers\Html;
-use <?= $generator->indexWidgetType === 'grid' ? "yii\\grid\\GridView" : "yii\\widgets\\ListView" ?>;
+<?php if ($generator->indexWidgetType === 'grid'): ?>
+use yii\helpers\Url;
+use yii\grid\ActionColumn;
+use yii\grid\GridView;
+use yii\grid\SerialColumn;
+<?php else: ?>
+use yii\widgets\ListView;
+<?php endif; ?>
+use <?= $generator->modelClass ?>;
 <?= $generator->enablePjax ? 'use yii\widgets\Pjax;' : '' ?>
 
-/* @var $this yii\web\View */
-<?= !empty($generator->searchModelClass) ? "/* @var \$searchModel " . ltrim($generator->searchModelClass, '\\') . " */\n" : '' ?>
-/* @var $dataProvider yii\data\ActiveDataProvider */
-
-$this->title = <?= $generator->generateString(Inflector::pluralize(Inflector::camel2words(StringHelper::basename($generator->modelClass)))) ?>;
+$this->title = <?= $generator->generateString(Inflector::pluralize(Inflector::camel2words($modelClass))) ?>;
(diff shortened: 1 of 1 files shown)


src/generators/crud/default/views/index.php
```diff
@@ -1,79 +1,104 @@
...
+    <p class="action-bar">
+        <?= '<?= ' ?>Html::a(<?= $generator->generateString('Create ' . $modelName) ?>, ['create'], ['class' => 'btn btn-success']) ?>
     </p>
 
-<?= $generator->enablePjax ? "    <?php Pjax::begin(); ?>\n" : '' ?>
+    <?= $generator->enablePjax ? '<?php Pjax::begin(); ?>' : '' ?>
+
 <?php if(!empty($generator->searchModelClass)): ?>
-<?= "    <?php " . ($generator->indexWidgetType === 'grid' ? "// " : "") ?>echo $this->render('_search', ['model' => $searchModel]); ?>
+    <?= '    <?php ' . ($generator->indexWidgetType === 'grid' ? '// ' : '') ?>echo $this->render('_search', ['model' => $searchModel]); ?>
```
samdark, line 51:
```suggestion
    <?= '<?php ' . ($generator->indexWidgetType === 'grid' ? '// ' : '') ?>echo $this->render('_search', ['model' => $searchModel]); ?>
```

src/generators/crud/default/views/index.php
```diff
@@ -1,79 +1,104 @@
...
+        'columns' => [
+            ['class' => SerialColumn::class<?= $php7 ? '' : 'Name()' ?>],
+            <?php
+            /** @var \yii\base\Model $model */
+            $model = \Yii::createObject($generator->modelClass);
+            $pk = $model instanceof \yii\db\ActiveRecord ? $model::primaryKey() : [];
+            $count = 1;
+            $tableSchema = $generator->getTableSchema();
+            $labels = $model->attributeLabels();
+            foreach ($model->attributes() as $attribute) {
```
samdark, line 67: The idea to use model attributes instead of table schema is alright 👍

WinterSilence: @samdark need rewrite generator, but I'm don't have time :(

src/generators/crud/default/views/index.php
```diff
@@ -1,79 +1,104 @@
...
-        }
-    }
-}
-?>
-
-            ['class' => 'yii\grid\ActionColumn'],
-        ],
+        <?= $generator->searchModelClass ? "'filterModel' => \$searchModel,\n" : ''; ?>
+        'columns' => [
+            ['class' => SerialColumn::class<?= $php7 ? '' : 'Name()' ?>],
```
machour, line 59: I think we should stick to generating ::className(), as the resulting code needs to be PHP 5.4 compatible.

I could be working with PHP 7 locally and still deploy to PHP 5.4

src/generators/crud/default/views/index.php
```diff
@@ -1,79 +1,104 @@
...
+                    $format = $generator->generateColumnFormat($column);
+                }
+                $label = isset($labels[$attribute]) ? $labels[$attribute] : Inflector::humanize($attribute);
+                echo sprintf("                %s'%s:%s:%s',\n", $count < 6 ? '' : '// ', $attribute, $format, $label);
+                $count++;
+            }
+            ?>
+            [
+                'class' => ActionColumn::class<?= $php7 ? '' : 'Name()' ?>,
+                'urlCreator' => static function ($action, <?= $modelClass ?> $model/*, $key, $index, $column */) {
```
machour, line 83: What's the motivation behind this change ?

machour (changes requested):
Great idea !

Would you please post an example output in comment ?

machour:
@samdark shouldn't we have tests in this extension, including & comparing outputs ? That would make it easier to track changes

samdark:
@machour yeah but we don't have any :)

src/generators/crud/default/views/index.php
machour: Related: https://github.com/yiisoft/yii2-gii/issues/437

We need to add a global option like `--php-version` and do what you did, but that needs to be handled in a different PR if you're up for it

WinterSilence: @machour  approve your initiative

src/generators/crud/default/views/index.php
samdark: @WinterSilence ideally that should be done separately and merged right away...

rob006:
> yeah but we don't have any :)

Maybe we should start with adding these, and only then start refactoring this extension. Right now reviewing such PRs is a nightmare - it is often  really hard to understand what are the actual implications of some changes.

PR #476, opened 2021-10-12
Remove old datalist options in view `generators/model/form`

WinterSilence:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes
| Fixed issues  |


WinterSilence:
@samdark
1. create table `x`
2. generate model for table `x`
3. remove/rename table `x` or switch database
4. check table list - table `x` still here

PR #483, opened 2021-11-05
Normalize action method in DefaultController::actionAction()

WinterSilence:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |


WinterSilence:
@samdark 
> Given an action named "xyz", the method "action**X**yz()" in the generator will be called.

rob006:
It does not matter, since method names are case insensitive in PHP.

WinterSilence:
@rob006 yes, captain...

PR #488, opened 2021-11-10
Fix #482 in index template of CRUD generator

### LINKED ISSUE #482: CRUD generator template issue - view/update/delete raises error

1. Create the SQL database 'world' with this script: https://www.roc.ovh/attachments/53
2. You pick the 'country' table from the SQL file and create the table 'country'.
3. Start a new Yii project, create a model of the table 'country' using the model creator.
4. Start the Yii crud generator and create a crud.
5. Go to 'country' (grid view) and navigate to the view (little eye-icon on the right side).
6. (Yii) Smart routing is used - I don't think this really matters?

### What is the expected result?

I expect to see the view in which the current record ('country') is shown.

### What do you get instead?

Error: Bad request (#400) Missing required parameters: Code 

The problem is that the URL is http://localhost:8080/country/view?id=ABW
and the code expects 'Code' as parameter: public function actionView($Code)

http://localhost:8080/country/view?Code=ABW
solves one step but that the code stops at line 127 of the 'CountryController' where the 'id' is expected.

Using an older version of the CRUD templates, f.e. version 2.0.37 solves the issue.

### Additional info

Issue occurs when using navigating to the view, update and delete from the grid view, using the small icons.
Issue also occurs when navigating straight to a view, f.e. http://localhost:8080/country/view?Code=ABW 

I need more testing, but I think the issue does occur if the primary key of a table is *not* named 'id'.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.43
| PHP version      | 7.2.34
| Operating system |Win10

WinterSilence:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #482

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Yii Framework 2 gii extension Change Log
 - Bug #467: Fix view `generators/crud/default/controller` (WinterSilence, cjrf)
 - Bug #476: Fix stucking datalist options in form generator (WinterSilence)
 - Bug #486: Update `assets/js/bs4-native.min.js` to latest version (WinterSilence)
+- Bug #488: Fix `ActionColumn::$urlCreator` in index template of CRUD generator (WinterSilence)
 
 
 2.2.3 August 09, 2021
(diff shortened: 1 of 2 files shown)


samdark (approved):
Please add a line for changelog. Thanks.

WinterSilence:
@samdark done

samdark:
👍

Issue #400, opened 2019-02-14
change model generator to use class instead of className()


Isitar:
Hello

I would appreciate if gii uses ::class instead of className() in templates. 
Is there any reason this isn't done? className() is depricated and slower than ::class.

Example-Line to be changed in model generator for relations
https://github.com/yiisoft/yii2-gii/blob/bfe6044f7ee8089923c0bbfda7394a94925c4b0e/src/generators/model/Generator.php#L533

alexkart:
Hello, maybe it's done for compatibility with PHP 5.4?

machour:
`::class` was implemented in PHP 5.5 but Yii2 tries to be compatible with PHP 5.4, so I don't think we will change that. @samdark ?

machour:
We could consider dropping PHP 5.X for 2.1, as we did for swiftmailer

samdark:
I don't think it worth doing in 2.x. For 3.x — definitely.

Issue #356, opened 2018-04-27
Replace deprecated ::className() with ::class


Deele:
Since 2.0.14. On PHP >=5.5, use `::class` instead of `::className()`.

See https://github.com/yiisoft/yii2/blob/master/framework/base/BaseObject.php#L82

At least 11 results https://github.com/yiisoft/yii2-gii/search?q=::className()

samdark:
Yes, in the next major version that is for Yii 2.1 we can do that.

desenvolvedorindie:
Probabily need check version of php like this

version_compare(phpversion(),'5.5.0')

To keep compatibility

samdark:
No. The change should be done in https://github.com/yiisoft/yii-gii instead.

sokollondon:
More than a year has passed. Problem still exists. PHPStorm inspection shows warning!
Help )

schmunk42:
@sokollondon Yii2 is compatible with PHP 5.4 but `:class` is only available from 5.5 - so this won't happen it Yii2.

machour:
Since Yii2 is still going to be used for a long time, I'm wondering if it's worth providing an option to generate PHP7 code.. 🤔

schmunk42:
> Since Yii2 is still going to be used for a long time, I'm wondering if it's worth providing an option to generate PHP7 code..

Yeah, it's a valid point. Is there more than just `:class`? This could be done with an option.

Deele:
> 
> 
> > Since Yii2 is still going to be used for a long time, I'm wondering if it's worth providing an option to generate PHP7 code..
> 
> Yeah, it's a valid point. Is there more than just `:class`? This could be done with an option.

That could be done with just another template bundled together with core Gii, I approve.

rob006:
I'm not sure if we need this as an option. We could just release new major release (2.2) which will generate code according to current standards (everything before PHP 7.1 is already EOLed) and everyone who stuck in PHP 5 age could use old releases of Gii (2.0 or 2.1).

albertvangink:
Can i Find "className()" & Replace them all with "class" safe at a fresh install?

samdark:
Likely yes.

albertvangink:
There were a few files that didn't like it :-(

C:\xampp\htdocs\basic\vendor\sebastian\type\src\ObjectType.php

C:\xampp\htdocs\basic\vendor\yiisoft\yii2-gii\src\generators\model\Generator.php

[...]

PR #342, opened 2018-03-02
change signature to ::class

bscheshirwork:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | yes
| Tests pass?   | yes/no

related to `yii 2.1` and prepare for it.

PR #305, opened 2017-09-18
[2.1] changed className() to class

schmunk42:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | ?

PR #489, opened 2021-11-14
Added support for the `::class` constant in model generator.

rhertogh:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #305, #342, #356, #400, #403, #427, #437, #493

This PR adds the `useClassConstant` option to the model generator. When `true` `::class` instead of `::className()` is used.

~~At the moment the default value for `useClassNameResolutionConstant` is `false`. We could decide to set it to `true` since only users on PHP 5.4 would need the old method.~~
Update 2022-06-15:
The default value for `useClassConstant` is `null` which means it will be automatically resolved to `true` if the PHP version >= 5.5 and to `false` ortherwise. Fixing it to either `true` or `false` is also possible.

### DIFF
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -50,7 +50,7 @@ class Generator extends \yii\gii\Generator
     public $baseClass = 'yii\db\ActiveRecord';
     public $generateRelations = self::RELATIONS_ALL;
     public $generateJunctionRelationMode = self::JUNCTION_RELATION_VIA_TABLE;
-    public $useClassNameResolutionConstant = false;
+    public $useClassNameResolutionConstant = null;
     public $generateRelationsFromCurrentSchema = true;
     public $generateLabelsFromComments = false;
     public $useTablePrefix = false;
@@ -76,6 +76,15 @@ class Generator extends \yii\gii\Generator
      * @var string[]
      */
     protected $classNames = [];
+
+    public function init()
+    {
+        parent::init();
+
+        if ($this->useClassNameResolutionConstant === null) {
+            $this->useClassNameResolutionConstant = version_compare(PHP_VERSION, '5.5') >= 0;
+        }
+    }
 
     /**
      * @inheritdoc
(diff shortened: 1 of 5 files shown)


WinterSilence:
not so good, but still better than nothing 👍🏻

src/generators/model/Generator.php
```diff
@@ -50,6 +50,7 @@ class Generator extends \yii\gii\Generator
     public $baseClass = 'yii\db\ActiveRecord';
     public $generateRelations = self::RELATIONS_ALL;
     public $generateJunctionRelationMode = self::JUNCTION_RELATION_VIA_TABLE;
+    public $useClassNameResolutionConstant = null;
```
bizley, line 53: I would go with `useClassConstant` because it's shorter.

src/generators/model/Generator.php
```diff
@@ -76,6 +77,14 @@ class Generator extends \yii\gii\Generator
      */
     protected $classNames = [];
 
+    public function init()
+    {
+        parent::init();
+
+        if ($this->useClassNameResolutionConstant === null) {
+            $this->useClassNameResolutionConstant = version_compare(PHP_VERSION, '5.5') >= 0;
```
bizley, line 85:
```suggestion
            $this->useClassNameResolutionConstant = PHP_VERSION_ID >= 50500;
```

src/generators/model/Generator.php
```diff
@@ -1146,4 +1164,20 @@ protected function isColumnAutoIncremental($table, $columns)
...
+    /**
+     * Returns the class name resolution
+     * @param string $class
+     * @return string
+     * @see $useClassNameResolutionConstant
+     * @since 2.2.5
+     */
+    protected function generateClassNameResolution($class)
+    {
+        if ($this->useClassNameResolutionConstant) {
```
WinterSilence, line 1177: `return $class . '::class' . (PHP_VERSION_ID < 50500 ? 'Name()' : '');`

bizley: Shouldn't it depend on the flag?

WinterSilence: @bizley i mean property no need because used only in current method

bizley: But isn't it configurable? I'm confused a bit...

WinterSilence: @bizley I missed this moment

rhertogh: @bizley Yes, it should indeed be configurable (in case the PHP version at generation differs from the targeted PHP version).
@WinterSilence Suggested code optimization has been implemented

rhertogh:
PR is updated based on feedback.

samdark (approved):

samdark:
Thank you.

PR #500, opened 2022-04-12
Added use statement for model namespace

mohamed-nazim:
php 7.4 is giving errors about missing namespace use statement because the 'urlCreator' closure uses '$modelClass' that doesn't have a namespace, and results in 'Argument 2 passed to yii\base\View::{closure}() must be an instance of ModelClassName'

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


samdark (approved):
Please add a line for CHANGELOG. Thanks.

samdark (approved):

PR #499, opened 2022-04-12
Added support for field types 'date', 'datetime' & 'time'

mohamed-nazim:
Added input field of type 'date', 'datetime' & 'time' in form/search fields.

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

PR #501, opened 2022-04-15
Date time fields support in crud form

mohamed-nazim:
Updated my branch name for the previous [PR](https://github.com/yiisoft/yii2-gii/pull/499) and it ended up closed. So resubmitting.

This PR adds support for input field of type 'date', 'datetime' & 'time' in form/search fields and is based on current browser standards for date / date time input field. MDN [definition](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) for `date` says it uses `YYYY-MM-DD` standard and the [definition](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local) for `datetime-local` uses `YYYY-MM-DDT00:00` standard - which is what is sent over when the form is submitted (although what is rendered on the screen is based on local device settings/locale).

| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️

### DIFF
--- a/src/generators/crud/Generator.php
+++ b/src/generators/crud/Generator.php
@@ -275,6 +275,10 @@ class Generator extends \yii\gii\Generator
             return "\$form->field(\$model, '$attribute')->textInput(['type' => 'datetime-local'])";
         }
 
+        if ($column->type === Schema::TYPE_TIME) {
+            return "\$form->field(\$model, '$attribute')->textInput(['type' => 'time'])";
+        }
+
         if ($column->type === 'text') {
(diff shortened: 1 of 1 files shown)


WinterSilence:
it's same to prev PR

mohamed-nazim:
> it's same to prev PR
> 
> > where's your browsers and databases tests?

Yes @WinterSilence it's same. It got closed, because I messed up my branch name. See ☝️I said so at the beginning :)

browsers and databases tests - shall do and post asap.

PR #511, opened 2023-01-29
Fix validation for `messageCategory`

rob006:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

Previous regexp matched strings like `<div>`.


bizley:
Thanks

PR #513, opened 2023-01-29
Update regexp for `messageCategory` validation

rob006:
Appendix for https://github.com/yiisoft/yii2-gii/pull/511. While it fixed original intentions from https://github.com/yiisoft/yii2-gii/pull/456, it seem to be too restrictive. Even documentation proposes categories like `app/error`, so it does not make sense to disallow `/` character. In addition to that `.` and `-` could be also useful for message categories.


bizley:
👍🏻

PR #514, opened 2023-01-29
Fix tests

rob006:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -109,7 +109,7 @@ class Generator extends \yii\gii\Generator
     public function rules()
     {
         return array_merge(parent::rules(), [
-            [['db', 'tableName', 'modelClass', 'baseClass', 'queryClass', 'queryBaseClass'], 'trim', 'skipOnEmpty' => true],
+            [['db', 'tableName', 'modelClass', 'baseClass', 'queryClass', 'queryBaseClass'], 'trim'],
             [
                 ['ns', 'queryNs'],
                 'filter',
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -39,18 +39,18 @@ class Generator extends \yii\gii\Generator
     /**
      * @var string
      */
-    public $tableName;
+    public $tableName = '';
     /**
      * @var string
      */
-    public $modelClass;
+    public $modelClass = '';
     /**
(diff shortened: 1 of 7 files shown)


src/generators/model/Generator.php
```diff
@@ -1072,8 +1072,8 @@ protected function generateClassName($tableName, $useSchemaName = null)
         $patterns = [];
         $patterns[] = "/^{$db->tablePrefix}(.*?)$/";
         $patterns[] = "/^(.*?){$db->tablePrefix}$/";
-        if (strpos($this->tableName, '*') !== false) {
-            $pattern = $this->tableName;
+        if (strpos($tableName, '*') !== false) {
+            $pattern = $tableName;
```
rob006, line 1076: I'm not really sure about this change, but it was weird that method ignored passed argument and used property instead. There are no failing tests after this change, so it could just be a bug.

bizley: This was like that from the beginning. $tableName is a proper name of the table while $this->tableName is provided by a user and can contain asterisk.

src/generators/model/Generator.php
```diff
@@ -109,7 +109,7 @@ public function getDescription()
     public function rules()
     {
         return array_merge(parent::rules(), [
-            [['db', 'tableName', 'modelClass', 'baseClass', 'queryClass', 'queryBaseClass'], 'filter', 'filter' => 'trim', 'skipOnEmpty' => true],
```
samdark, line 112: Is it necessay to remove `skipOnEmpty`?

rob006: Maybe not, but `trim` validator ensures that value is string after validation (it could be `null` without it), and there is no `skipOnEmpty` setting in other places.

src/generators/model/Generator.php
rob006: I'll revert it then. But lack of tests for this is a little worrying.

rob006: Done.

bizley (approved):

samdark (approved):

samdark:
Thank you!

PR #537, opened 2023-09-27
Generating rules for the fields with default values

manky:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

This may sound unimportant, but if you are using **yii\behaviors\AttributeTypecastBehavior** and the cast depends on the **rules()** method of your model, for example, fields validated via **yii\validators\NumberValidator** and without any value will be cast to integer/float. So the value of the empty fields will become **0** 

_An important part is that default rules should be at the beginning of the rule list._

### DIFF
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -438,12 +438,21 @@ class Generator extends \yii\gii\Generator
     {
         $types = [];
         $lengths = [];
+        $nullable = [];
+        $defaultValues = [];
         foreach ($table->columns as $column) {
             if ($column->autoIncrement) {
                 continue;
             }
             if (!$column->allowNull && $column->defaultValue === null) {
                 $types['required'][] = $column->name;
+            } elseif ($column->allowNull && $column->defaultValue === null) {
+                $nullable[] = $column->name;
+            } elseif (is_scalar($column->defaultValue)) {
+                if (array_key_exists($column->defaultValue, $defaultValues)) {
+                    $defaultValues[$column->defaultValue] = [];
+                }
+                $defaultValues[$column->defaultValue][] = $column->name;
             }
             switch ($column->type) {
                 case Schema::TYPE_SMALLINT:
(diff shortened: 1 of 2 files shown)


samdark (approved):
Looks good. Please add a line for changelog. Thanks.

bizley (approved):

samdark:
Thank you!

src/generators/model/Generator.php
```diff
@@ -438,12 +438,21 @@ public function generateRules($table)
...
         foreach ($table->columns as $column) {
             if ($column->autoIncrement) {
                 continue;
             }
             if (!$column->allowNull && $column->defaultValue === null) {
                 $types['required'][] = $column->name;
+            } elseif ($column->allowNull && $column->defaultValue === null) {
+                $nullable[] = $column->name;
+            } elseif (is_scalar($column->defaultValue)) {
+                if (array_key_exists($column->defaultValue, $defaultValues)) {
```
xcopy, line 452: I believe this should be `!array_key_exists(...)`.

Otherwise,

```php
$defaultValues[$column->defaultValue] = [];
```

will reset that array in the next iteration, right?

xcopy: In my case the expected result is

```php
[['working_hours', 'payable_hours', 'payable_leave_days', 'non_payable_leave_days', 'business_trip_days', 'sick_leave_days', 'day_off_days'], 'default', 'value' => 0],
```

and the actual one is:

```php
[['working_hours'], 'default', 'value' => 0],
```

src/generators/model/Generator.php
xcopy, line 1: I've got 

```php
[[...], 'default', 'value' => null],
```

after 

```php
[[...], 'default', 'value' => 0],
```

have been generated, i.e. because of

```php
if ($driverName === 'pgsql' && $type === 'integer') {
    $rules[] = "[['" . implode("', '", $columns) . "'], 'default', 'value' => null]";
}
```

i.e. why do I need `default = null` if the `default = 0` is enough for me...

xcopy:
@manky @samdark guys anyone?

PR #542, opened 2023-10-17
Use the name of the table to create the relation

thiagotalma:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

Use the real name of the table to create the relationship when there is a composite primary key.

Prevents the generation of random relationship names.

**Turn this:**

```PHP
<?php

(...)

/**
 *
 * @property int $id
 * @property int $employee_id
 *
 * @property Employee $employee
 * @property Route $employee0 <---
 */
class Company extends \yii\db\ActiveRecord
{
    (...)

    /**
     * Gets query for [[Employee]].
     *
     * @return \yii\db\ActiveQuery
     */
    public function getEmployee()
    {
        return $this->hasOne(Employee::class, ['id' => 'employee_id']);
    }

    /**
     * Gets query for [[Employee0]]. <---
     *
     * @return \yii\db\ActiveQuery
     */
    public function getEmployee0() /* <--- */
    {
        return $this->hasOne(Route::class, ['employee_id' => 'employee_id', 'id' => 'route_id']);
    }
}
```

**Into this:**

```PHP
<?php

(...)

/**
 *
 * @property int $id
 * @property int $employee_id
 * @property string $route_id
 *
 * @property Employee $employee
 * @property Route $route <---
 */
class Company extends \yii\db\ActiveRecord
{
    (...)

    /**
     * Gets query for [[Employee]].
     *
     * @return \yii\db\ActiveQuery
     */
    public function getEmployee()
    {
        return $this->hasOne(Employee::class, ['id' => 'employee_id']);
    }

    /**
     * Gets query for [[Route]]. <---
     *
     * @return \yii\db\ActiveQuery
     */
    public function getRoute() /* <--- */
    {
        return $this->hasOne(Route::class, ['employee_id' => 'employee_id', 'id' => 'route_id']);
    }
}
```

### DIFF
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -52,6 +52,7 @@ class Generator extends \yii\gii\Generator
     public $generateJunctionRelationMode = self::JUNCTION_RELATION_VIA_TABLE;
     public $useClassConstant;
     public $generateRelationsFromCurrentSchema = true;
+    public $generateRelationNameFromDestinationTable = false;
     public $generateLabelsFromComments = false;
     public $useTablePrefix = false;
     public $standardizeCapitals = false;
@@ -130,7 +131,7 @@ class Generator extends \yii\gii\Generator
             [['generateRelations'], 'in', 'range' => [self::RELATIONS_NONE, self::RELATIONS_ALL, self::RELATIONS_ALL_INVERSE]],
             [['generateJunctionRelationMode'], 'in', 'range' => [self::JUNCTION_RELATION_VIA_TABLE, self::JUNCTION_RELATION_VIA_MODEL]],
             [
-                ['generateLabelsFromComments', 'useTablePrefix', 'useSchemaName', 'generateQuery', 'generateRelationsFromCurrentSchema', 'useClassConstant', 'enableI18N', 'standardizeCapitals', 'singularize'],
+                ['generateLabelsFromComments', 'useTablePrefix', 'useSchemaName', 'generateQuery', 'generateRelationsFromCurrentSchema', 'generateRelationNameFromDestinationTable', 'useClassConstant', 'enableI18N', 'standardizeCapitals', 'singularize'],
                 'boolean'
             ],
             [['messageCategory'], 'validateMessageCategory', 'skipOnEmpty' => false],
@@ -153,6 +154,7 @@ class Generator extends \yii\gii\Generator
             'generateRelations' => 'Generate Relations',
             'generateJunctionRelationMode' => 'Generate Junction Relations As',
             'generateRelationsFromCurrentSchema' => 'Generate Relations from Current Schema',
+            'generateRelationNameFromDestinationTable' => 'Generate Relations Name Using Destination Table Name',
             'useClassConstant' => 'Use `::class`',
             'generateLabelsFromComments' => 'Generate Labels from DB Comments',
             'generateQuery' => 'Generate ActiveQuery',
@@ -195,6 +197,7 @@ class Generator extends \yii\gii\Generator
                 Make sure you also generate the junction models when using the "Via Model" option.
             ',
             'generateRelationsFromCurrentSchema' => 'This indicates whether the generator should generate relations from current schema or from all available schemas.',
+            'generateRelationNameFromDestinationTable' => 'This indicates whether the relation name should reflect the target table name.',
             'useClassConstant' => 'Use the `::class` constant instead of the `::className()` method.',
             'generateLabelsFromComments' => 'This indicates whether the generator should generate attribute labels
                 by using the comments of the corresponding DB columns.',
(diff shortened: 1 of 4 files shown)


samdark:
What if both relations are needed?

thiagotalma:
Look, I'm not eliminating the relationship.

Just renaming. Using the table name instead of the primary key.

samdark:
Tests should be fixed.

thiagotalma:
> Tests should be fixed.

@samdark Done!

xepozz (approved):

thiagotalma:
@samdark Is there any chance this change will be present in the next release?

xepozz:
@bizley @terabytesoftw Could you please review?

terabytesoftw (approved):

rhertogh:
I think there should be a configuration option for this change for backwards compatibility. E.g. when used with https://github.com/schmunk42/yii2-giiant.

bizley (approved):

bizley:
I would not want to keep the old way. @schmunk42 could you check if this change is problematic with your package?

schmunk42:
One thing is, that the workflow with `giiant` is to regenerate base-models everytime your db changes, which would be problematic here, since the relations would change.

But another more important thing is, that if you have a table `company` with 2 FKs to Employee eg. `president_id` and `vice_president_id`, and you take the relation name from the table name you'll end up with `getEmployee` and `getEmployee0`. In this use-case you are better off with relation names from the column.

I'd not change this. At least not for yii2.
Or make it configurable.

bizley:
Ok, I see. If we correct this in one case, the other one is broken. @thiagotalma could you make it configurable?

thiagotalma:
Configuration in the module or on the gii screen?

schmunk42:
gii screen I think

src/generators/model/Generator.php
```diff
@@ -153,6 +154,7 @@ public function attributeLabels()
             'generateRelations' => 'Generate Relations',
             'generateJunctionRelationMode' => 'Generate Junction Relations As',
             'generateRelationsFromCurrentSchema' => 'Generate Relations from Current Schema',
+            'generateRelationNameFromDestinationTable' => 'Generate Relations Name Using Destination Table Name',
```
thiagotalma, line 157: I'll need help with the variable name and the hint.

I am not fluent in English.

bizley: Maybe `Generate Relation Names Using Target Table Name`

schmunk42:
@thiagotalma There should be tests for both cases (original ones and the new one).

thiagotalma:
Can anyone help me with the tests, please?
(Yes, it's a shame. I know)

schmunk42:
You might need an additional param in your data provider:

```
public function testRelations($tableName, $fileName, $useClassConstant, $relations, $fromDestTable = false)
    {
        $generator = new ModelGenerator();
...
        $generator->generateRelationNameFromDestinationTable = $fromDestTable;
```

And use it in your new use-cases in `relationsProvider()`:
```
            ['product_language', 'ProductLanguage.php', false, [
                [
                    'name' => 'function getSupplier()',
                    'name' => 'function getProduct()',
                    'relation' => "\$this->hasOne(Product::className(), ['supplier_id' => 'supplier_id', 'id' => 'id']);",
                    'expected' => true,
                ],
                [
                    'name' => 'function getSupplier0()',
                    'name' => 'function getSupplier()',
                    'relation' => "\$this->hasOne(Supplier::className(), ['id' => 'supplier_id']);",
                    'expected' => true,
                ],
            ],
         true // $fromDestTable
         ],
``` 

The existing ones should be kept/reverted.

rhertogh:
> But another more important thing is, that if you have a table `company` with 2 FKs to Employee eg. `president_id` and `vice_president_id`, and you take the relation name from the table name you'll end up with `getEmployee` and `getEmployee0`. In this use-case you are better off with relation names from the column.
> 
> I'd not change this. At least not for yii2. Or make it configurable.

I was thinking about this some more since I encountered a similar situation today.
In my opinion the best option would be to use the name of the foreign key itself.
In the example from @schmunk42 above the foreign keys probably would be named "president" and "vice_president" respectively.

So my proposal would be to make it indeed configurable as discussed above and add a third option.
The configuration would be `$generateRelationNameFrom`, with the options:
- column name
- target table name
- foreign key name

thiagotalma:
Done!

Thanks, @schmunk42

schmunk42:
@thiagotalma You're welcome.

LGTM now :)

bizley (approved):

samdark (approved):

samdark:
Thank you!

bizley:
Oh, indeed. @thiagotalma could you prepare another PR correcting this?

thiagotalma:
@bizley 
Done!
https://github.com/yiisoft/yii2-gii/pull/544

thiagotalma:
Hello guys!

What do you think about releasing the new version?
The milestone is already cleared.

I'm just waiting for these changes to move forward with my project.

@samdark

PR #544, opened 2023-11-21
Use the name of the table to create the relation

thiagotalma:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

Use the real name of the table to create the relationship when there is a composite primary key.

Prevents the generation of random relationship names.

Turn this:

```PHP
<?php

(...)

/**
 *
 * @property int $id
 * @property int $employee_id
 *
 * @property Employee $employee
 * @property Route $employee0 <---
 */
class Company extends \yii\db\ActiveRecord
{
    (...)

    /**
     * Gets query for [[Employee]].
     *
     * @return \yii\db\ActiveQuery
     */
    public function getEmployee()
    {
        return $this->hasOne(Employee::class, ['id' => 'employee_id']);
    }

    /**
     * Gets query for [[Employee0]]. <---
     *
     * @return \yii\db\ActiveQuery
     */
    public function getEmployee0() /* <--- */
    {
        return $this->hasOne(Route::class, ['employee_id' => 'employee_id', 'id' => 'route_id']);
    }
}
```

Into this:

```PHP
<?php

(...)

/**
 *
 * @property int $id
 * @property int $employee_id
 * @property string $route_id
 *
 * @property Employee $employee
 * @property Route $route <---
 */
class Company extends \yii\db\ActiveRecord
{
    (...)

    /**
     * Gets query for [[Employee]].
     *
     * @return \yii\db\ActiveQuery
     */
    public function getEmployee()
    {
        return $this->hasOne(Employee::class, ['id' => 'employee_id']);
    }

    /**
     * Gets query for [[Route]]. <---
     *
     * @return \yii\db\ActiveQuery
     */
    public function getRoute() /* <--- */
    {
        return $this->hasOne(Route::class, ['employee_id' => 'employee_id', 'id' => 'route_id']);
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Yii Framework 2 gii extension Change Log
 
 - Bug #532: Return `ExitCode::USAGE` on command input validation error (egmsystems)
 - Enh #537: Generating rules for the fields with default values (manky)
+- Enh #542: Use the table name to create the relation (thiagotalma)
 
 
 2.2.6 May 22, 2023
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Yii Framework 2 gii extension Change Log
 
 - Bug #532: Return `ExitCode::USAGE` on command input validation error (egmsystems)
 - Enh #537: Generating rules for the fields with default values (manky)
-- Enh #543: Use the table name to create the relation (thiagotalma)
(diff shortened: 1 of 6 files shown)


bizley:
Since this just brings back the removed code please don't update the changelog.

bizley:
Possibly, we need to check the current milestone state.

bizley (approved):

PR #545, opened 2023-11-22
Update Generator.php

spzgy:
In yii\db\BaseActiveRecord, there is no getTableSchema method. When using Gii to generate CRUD for non-relational databases, an error occurs because the getTableSchema method cannot be found.

| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ✔️/❌
| Breaks BC?    | ✔️/❌
| Fixed issues  |


bizley (approved):

samdark:
Please add a line for CHANGELOG. Thanks.

PR #546, opened 2024-02-10
Added return types, removed unused Yii from model.php, replaced imports

Erkinbek:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | Added return types

### DIFF
--- a/src/generators/crud/default/controller.php
+++ b/src/generators/crud/default/controller.php
@@ -38,6 +38,7 @@ use yii\data\ActiveDataProvider;
 use <?= ltrim($generator->baseControllerClass, '\\') ?>;
 use yii\web\NotFoundHttpException;
 use yii\filters\VerbFilter;
+use yii\web\Response;
 
 /**
  * <?= $controllerClass ?> implements the CRUD actions for <?= $modelClass ?> model.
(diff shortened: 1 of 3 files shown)


samdark:
It won't work with older PHP versions.

uldisn:
> It won't work with older PHP versions.

If create  own templates for each PHP version, it will work.

Erkinbek:
All time when i generate code with gii, my idea (PHPStorm) will shows this lines as not optimized code. So if people are using old versions of php, just need close this PR. Thank you for your attention :)

uldisn:
Idea no bad! In https://github.com/yiisoft/yii2-gii/blob/master/src/generators/model/Generator.php#L313 
```
if (phpversion() < '5.2' ) {
    $template = 'model_52.php';
} elseif {phpversion() >= '5.2' &&  phpversion() < '5.7'} {
   $template = 'model_52_57.php';
} elseif {phpversion() >= '5.7' &&  phpversion() < '7.4'} {
   $template = 'model_57_74.php';
} {
   $template = 'model_74.php';
}
```

For each PHP version own model_*php.

Erkinbek:
OK. I'll fix it

PR #547, opened 2024-02-12
Added return types for methods in models

Erkinbek:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  |❌
| Breaks BC?    | ❌
| Fixed issues  | Added method return types

### DIFF
--- a/src/generators/model/Generator.php
+++ b/src/generators/model/Generator.php
@@ -310,9 +310,12 @@ class Generator extends \yii\gii\Generator
                 'relationsClassHints' => $this->generateRelationsClassHints($tableRelations, $this->generateQuery),
                 'enum' => $this->getEnum($tableSchema->columns),
             ];
+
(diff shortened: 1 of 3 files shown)


src/generators/model/Generator.php
```diff
@@ -310,9 +310,12 @@ public function generate()
                 'relationsClassHints' => $this->generateRelationsClassHints($tableRelations, $this->generateQuery),
                 'enum' => $this->getEnum($tableSchema->columns),
             ];
+
+            $template = phpversion() < '7' ? 'model.php': 'model_from_7.php';
```
samdark, line 314:
```suggestion
            $template = phpversion() < '7' ? 'model.php': 'model_with_types.php';
```

samdark:
Overall it's a good idea.

samdark:
@Erkinbek do you have time to fix the naming?

Issue #551, opened 2024-12-25
Add form name to search model


evil1:
When generate CRUD with gii the default search model template generates pretty decent code, however it would be nice to be able to pass the `$formName` parameter into search method. This is great, when using SearchModel together with API, and attribute received in the request are not in `SearchModelClass[attibuteName]` format.

| Q                | A
| ---------------- | ---
| Yii version      | 2.0.51
| PHP version      | 8.3.11
| Database version | pgsql
| Operating system | macOS Sequoia

PR #552, opened 2024-12-25
New feature for the issue #551

evil1:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #551

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ Yii Framework 2 gii extension Change Log
 
 2.2.7 under development
 -----------------------
-- Bug #551: Add optional parameter `$formName` to `SearchModel::search()` method (evil1)
+- Enh #551: Add optional parameter `$formName` to `SearchModel::search()` method (evil1)
 - Bug #531: Fix `yii\gii\console\GenerateAction` to use `stdout()` instead of echoing the output (egmsystems)
 - Bug #532: Return `ExitCode::USAGE` on command input validation error (egmsystems)
 - Enh #537: Generating rules for the fields with default values (manky)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ Yii Framework 2 gii extension Change Log
 
 2.2.7 under development
 -----------------------
-
+- Bug #551: Add optional parameter `$formName` to `SearchModel::search()` method (evil1)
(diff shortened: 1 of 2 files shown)


samdark (approved):
Please add a line to CHANGELOG. Thanks.

evil1:
> Please add a line to CHANGELOG. Thanks.

Done

samdark:
Forgot to push?

evil1:
> Forgot to push?

Added changelog to another issue, related to `yiisoft/db` :)

samdark:
Thank you!

rob006:
I don't get it. By default `formName()` is used as form name in `load()` and while generating inputs. After this change form name is specified explicitly for `load()`, but for inputs generation value from `formName()` is still used. So if you want to change form name you need to change it in two places now?

samdark:
@evil1 ?

evil1:
> I don't get it. By default `formName()` is used as form name in `load()` and while generating inputs. After this change form name is specified explicitly for `load()`, but for inputs generation value from `formName()` is still used. So if you want to change form name you need to change it in two places now?

`formName()` returns a class name (via `ReflectionClass::getShortName()` under the hood).
We are explicitly passing the same class name, to be able to pass custom form name (especially empty string) to the `search()` method.
Why do we have to change it somewhere else?

rob006:
> `formName()` returns a class name (via `ReflectionClass::getShortName()` under the hood).

`formName()` could be overwritten. Previously if I wanted to remove form name from search form, all I needed is to overwrite `formName()` to return empty string - new value was be respected both in view with search form and in `load()`. Now it is not enough - I also need to change default param value in `search()`, because `search()` ignores `formName()` after this PR.

So this solution is suboptimal for code that Gii generates, just to cover hypothetical (and IMO quite niche) use case.

evil1:
Got it. Should I revert this PR back?

rob006:
Changing method signature to `search($params, $formName = null)` should fix the problem I mentioned in https://github.com/yiisoft/yii2-gii/pull/552#issuecomment-2565213807. But I'm still not sure if this use case is common enough to have it in default template.

evil1:
Should I submit new PR to fix this issue?

samdark:
Yes. Please.

evil1:
> Yes. Please.

Added a PR [here](https://github.com/yiisoft/yii2-gii/pull/553). Set `$formName = null` as advised by @rob006

PR #553, opened 2025-01-05
Fix and issue introduced in PR #552

evil1:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | Fixe issue in PR #551 described by @rob006

### DIFF
--- a/src/generators/crud/default/search.php
+++ b/src/generators/crud/default/search.php
@@ -60,7 +60,7 @@ class <?= $searchModelClass ?> extends <?= isset($modelAlias) ? $modelAlias : $m
      *
      * @return ActiveDataProvider
      */
-    public function search($params)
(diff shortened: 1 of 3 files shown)


src/generators/crud/default/search.php
```diff
@@ -57,11 +57,11 @@ public function scenarios()
      * Creates data provider instance with search query applied
      *
      * @param array $params
-     * @param string $formName Form name to be used into `->load()` method. By default it is equal to the class name.
+     * @param string $formName Form name to be used into `->load()` method.
```
rob006, line 60:
```suggestion
     * @param string|null $formName Form name to be used into `->load()` method.
```

samdark:
Thank you!

PR #556, opened 2025-06-16
For names replace enum fields values symbols to abbreviations

uldisn:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | [(https://github.com/yiisoft/yii2-gii/issues/557)]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Yii Framework 2 gii extension Change Log
 2.2.8 under development
 -----------------------
 
+- Bug #558: Fix an issue with creating rules for multiple fields with the same default values (carono)
 - Bug #556: Non-alphabetical symbols in Enum values are replaced with their names when generating constants and function names.  (uldisn)
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Yii Framework 2 gii extension Change Log
 
 - Bug #558: Fix an issue with creating rules for multiple fields with the same default values (carono)
 
+
 2.2.7 February 13, 2025
 -----------------------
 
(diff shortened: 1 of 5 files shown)


schmunk42 (changes requested):
This breaks BC to my understanding!

What you could do is moving `constantName` to a callback with the current function as a default.

samdark (commented):
Need a line for CHANGELOG.

uldisn:
@samdark all generated code is covered

schmunk42:
Still breaks BC.

This [line](https://github.com/yiisoft/yii2-gii/blob/063ca8904336618b23786d9ddfa486e19d826dce/src/generators/model/Generator.php#L1223C17-L1223C96) should read:
```
$constantName = $this->constantNameCallback($column->name, $value)
```

This is the public property
```
public $constantNameCallback = function($name, $value){
  return strtoupper(Inflector::slug($name . ' ' . $value, '_'));
}
```

Somehow like that... I haven't tested it.

With this you should be able to plug-in any constant name generator function you want.

uldisn:
@schmunk42

schmunk42:
Honestly, this is a huge change, I can not tell if it is BC or not.
Also I am currently in holidays and have no time to review it.

I am against it, sorry.

uldisn:
?

schmunk42:
So, to my understanding here's some magic, which handles stuff like `A-Foo` the way it was.

https://github.com/yiisoft/yii2-gii/pull/556/files#diff-58ec64246f814497eb0524bc73fc58181b344edc0651e9e1014b004ee012554aR127-R145

But if you have something like `A (1)`:
- current code: `A_1` 
- PR code: `A_OPENING_PARENTHESIS_1_CLOSING_PARENTHESIS`

correct?

As mentioned before, it's too opinionated IMHO and an approach with a simple callback would be the best:
https://github.com/yiisoft/yii2-gii/pull/556#issuecomment-3063996336

PR #561, opened 2026-02-05
Apply Yii2 coding standards, raise min version to PHP 7.4

s1lver:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  |


terabytesoftw (approved):

.github/workflows/build.yml
mspirkov, line 1: Why isn't `yiisoft/yii2-actions/.github/workflows/phpunit.yml@master` being used here?

mspirkov: It seems that we can do something like this: https://github.com/yiisoft/yii2/blob/master/.github/workflows/ci-pgsql.yml

tests/generators/ModelGeneratorTest.php
```diff
@@ -411,7 +410,7 @@ public function testGenerateNotSingularizedClassNames()
     /**
      * @return array
      */
```
mspirkov: Now it's redundant

```suggestion
```

